PR #172, opened 2026-01-29
fix: distributed task routing with SetReturnNode

hyoklee:
## Summary
  - Fix distributed task response routing by adding `SetReturnNode()` call in
  HDF5 assimilator
  - Replace blocking `Wait()` with `co_await` to prevent worker thread deadlock
  - Add round-robin task distribution in scheduler to prevent deadlock when
  tasks spawn from containers
  - Add comprehensive debug logging for distributed task routing

  ## Changes
  - **hdf5_file_assimilator.cc**: Add `pool_query.SetReturnNode(current_node)`
  so remote nodes know where to send responses
  - **default_sched.cc**: Change from "current worker" to round-robin
  distribution among scheduler workers
  - **worker.cc, admin_runtime.cc, ipc_manager.h, core_client.h**: Add debug
  logging

  ## Test plan
  - [x] Tested with 4-node distributed processing (4 tiny datasets, 1 per node)
  - [x] All 4 nodes successfully process their assigned dataset
  - [x] All results correctly returned to orchestrator node
  - [x] Total completion time: 2.36 seconds for 4 datasets

### DIFF
--- a/context-assimilation-engine/core/src/factory/hdf5_file_assimilator.cc
+++ b/context-assimilation-engine/core/src/factory/hdf5_file_assimilator.cc
@@ -155,19 +155,14 @@ chi::TaskResume Hdf5FileAssimilator::Schedule(const AssimilationCtx& ctx, int& e
     std::vector<chi::Future<ProcessHdf5DatasetTask>> futures;
     futures.reserve(filtered_paths.size());
 
-    // Get current node ID for return routing
-    chi::u32 current_node = static_cast<chi::u32>(CHI_IPC->GetNodeId());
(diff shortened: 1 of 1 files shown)


lukemartinlogan (changes requested):
Please do not do PoolQuery::Physical + SetReturnNode. Please use PoolQuery::GetDIrectHash instead.

GetDirectHash() takes as input a container id, takes the modulo of that to the number of containers in the pool, and then translates the container id to a node id and sets SetReturnNode automatically. 

Let me know if that works

PR #315, opened 2026-02-28
fix: Compile errors related to CUDA & CTE in CPU mode

abscosmos:
This PR fixes issues preventing CTE from working when built with CUDA, and issues that were causing errors when building in CPU mode.


abscosmos:
This didn't introduce any new errors with asan; those tests were failing in 84f4e673c57c8cde6c6f846792925a0efd0c0650 also.

PR #338, opened 2026-03-05
Fix runtime health check to use hostfile hostname instead of localhost

hxu65:
The start/stop health checks were probing 127.0.0.1, but the runtime
  binds to the actual hostname from the hostfile (e.g., ares-comp-10).
  This caused a false 30s timeout warning on multi-node deployments.


hyoklee (approved):

lukemartinlogan (approved):
I didn't even now the jarvis package was doing this. We should just remove that. But approving for now.

PR #356, opened 2026-03-14
fix: Add missing Host/GPU guards

abscosmos:
Some functions/imports that are only for either host or GPU weren't guarded, or the guard wasn't defined correctly.


abscosmos:
Just merge #357, these commits are on there too.

PR #629, opened 2026-06-24
fix(runtime): tolerate stale replica responses in RecvOut + guard null run_ctx (#628)

### LINKED ISSUE #1: Implement migration

### LINKED ISSUE #628: Distributed: 3rd-replica send_map/net_key round-trip fails — cross-node data path hangs + leader SIGSEGV
## Summary

In a **3-node distributed cluster**, the cross-node `send_map` / `net_key` round-trip
fails for the **3rd replica** of a broadcast/replicated task. The forwarding node logs

```
RecvOutDeserialize [RecvOut] Task TaskId(... major:8 replica:2 ... net_key:N)
    FAILED: Origin task not found in send_map (size: 1) with net_key N
```

and the task never completes. This has two user-visible manifestations:

- **(A) Data-path hang.** A CTE `PutBlob` routed to the 3rd node (`PoolQuery::DirectHash(2)`)
  hangs indefinitely; the forwarding peer's connection to node 2 then drops.
- **(B) Leader SIGSEGV** (with SWIM enabled). The same `send_map`/origin-task mismatch in the
  dead-node **recovery** path dereferences a null container and crashes the leader:
  ```
  worker.cc:998  EndTask: container is null
  ipc_run2run.cc:462  RecvOutDeserialize Admin: origin_task has no RunContext
  <SIGSEGV>
  ```

A **2-node** cluster (2 replicas) works perfectly — including cross-node CTE `PutBlob`/`GetBlob`
with byte-verified integrity. The failure appears only when a **3rd replica** is added.

## Environment

3-node cluster, all x86-64 little-endian, **ZeroMQ over TCP** transport (Thallium/libfabric off,
io_uring off), port 9413, **`swim.enabled: false`** on all nodes (to take the crash path out):

| node | OS | clio-core commit |
|------|----|------------------|
| 0 | macOS 12 (Intel) | `b2a2049a` (HEAD) |
| 1 | CentOS 7 | `10f0904` |
| 2 | Windows 10 (rebuilt) | rebuilt from source |

The macOS↔Linux 2-node pool (commits `b2a2049a`↔`10f0904`) is fully functional; adding the
Windows 3rd node reproduces (A). (B) reproduces whenever any peer is reachable only one-way
(here: node-2 inbound 9413 was firewalled while its outbound succeeded), so SWIM flaps it
dead↔alive and the leader runs recovery.

## Steps to reproduce

1. Bring up a 3-node Chimaera cluster (shared hostfile, one bdev RAM tier + `clio_cte_core` per node,
   `targets.neighborhood: 3`).
2. From a client on node 1, `PutBlob` with `PoolQuery::DirectHash(2)` (route to node 2).
3. Observe: the call hangs; nodes 0 and 1 log `RecvOut ... replica:2 ... Origin task not found in send_map`.
4. For (B): enable SWIM and make node 2 reachable outbound-only (block inbound 9413). SWIM marks it
   dead → `TriggerRecovery` → a late task hits `EndTask: container is null` → SIGSEGV in the leader.

## Analysis

- `TaskId` (`context-runtime/include/clio_runtime/types.h:251`) is **all fixed-width** — six `u32`
  plus `size_t net_key_` — and `size_t` is 64-bit on Win64, macOS, and Linux, so this is **not**
  an LP64/LLP64 (`long`-width) wire mismatch.
- `net_key_` is **pointer-derived** (`reinterpret_cast<size_t>(task_ptr.ptr_)`,
  `ipc_cpu2cpu_zmq_impl.h:22`) and the receiver must echo it back **byte-identical** so the sender
  can match it in `send_map`. The error means the **3rd-replica response carries a net_key that does
  not match** the sender's pending entry — i.e. the broadcast/replicated `RecvOut` path drops or
  rewrites `net_key` for `replica_id == 2`, while it is preserved correctly for replicas 0/1.
- The 2-replica (2-node) path is unaffected, which points at the **≥3-replica broadcast response
  matching** in `ipc_run2run.cc` (`RecvOut*` / `RecvOutCompleteOriginTask`).

## Suggested fixes

1. **Stop the crash unconditionally:** in `worker.cc:998 EndTask`, if `container == nullptr`,
   log and drop the task — never dereference. (Robustness fix for (B).)
2. **Preserve `net_key` across replicated/broadcast responses** for all replicas, not just 0/1 —
   investigate why the `replica:2` response is unmatchable in `send_map` (likely in the
   `RecvOut` / `RecvOutCompleteOriginTask` round-trip).
3. **Dead-node recovery vs SWIM liveness race:** quarantine a late `SetAlive` for a node that is
   mid-recovery, and remove the node from the target set atomically before reorg, so a flapping /
   half-connected peer degrades to "node unusable" instead of crashing the leader.

Happy to provide full runtime logs from all three nodes and to test a patch.

hyoklee:
## What this addresses

Fixes #628 (3-node `send_map`/`net_key` round-trip: cross-node data-path hang + leader SIGSEGV).

This PR lands the **safe, verifiable robustness fixes** and documents the deeper routing/counting root cause (which needs the 3-node logs to confirm — see below). Note the issue's suggested fix #1 (`EndTask` container-null guard) is **already present** in `main` (and in `b2a2049a`), so this PR does the next layer.

## Changes

**1. `RecvOutDeserialize`: skip a stale replica response instead of aborting the whole batch** (`context-runtime/src/ipc/ipc_run2run.cc`)

`RecvOutDeserialize` returned non-zero when a `task_info`'s `net_key` was not in `send_map`, and `RecvOut` then bailed out *before* `RecvOutAggregate` ran. A response message can legitimately batch a live replica together with a straggler/duplicate whose origin was already completed and erased from `send_map` (the late 3rd-replica response, or one racing dead-node recovery). Aborting starved the sibling replicas in that message of deserialization, so the origin task never completed and the cross-node path hung. `RecvOutAggregate` already `continue`s on these exact conditions — this makes deserialize consistent so a single stale item only skips itself. Demoted the log to a warning since it is now an expected, tolerated case.

**2. `Worker::EndTask`: guard a null `RunContext`** (`context-runtime/src/worker.cc`)

`EndTask` dereferenced `run_ctx->container_` before null-checking `run_ctx`. A recovery/straggler task that lost its `RunContext` would SIGSEGV the worker. Now it is logged and dropped (mirrors the existing `container == nullptr` guard).

## Verification

Full clean rebuild + `ctest` via the Windows oneAPI/icx toolchain (`clio_d.r.bat`): **0 compiler errors**, both changed TUs compile, `clio_run.exe` links. These are strict robustness improvements — a found `net_key` still processes exactly as before, so the healthy 2-node path is unchanged.

I could **not** reproduce/verify the 3-node-specific behavior on a single node, so this PR intentionally does not blind-change the routing/counting logic below.

## Root-cause analysis for the remaining 3rd-replica failure (needs 3-node logs to confirm)

A per-replica `net_key` divergence is **impossible by construction**: `SendIn` stores one `send_map_[size_t(origin_task.ptr_)]` and every replica carries that identical `net_key` (`ipc_run2run.cc:167-170,206`). So "Origin task not found in send_map" means the entry was **erased early**, not that replica 2's key differs. Two concrete hazards explain that:

- **(A) Node-0 fallback collision in `PoolManager::GetContainerNodeId`** (`context-runtime/src/pool_manager.cc:738-740`). It returns `0` both for a *valid* node 0 and as the *"mapping not found"* sentinel — indistinguishable, since node ids are 0-based. `address_map_` is `.clear()`ed and rebuilt under a write lock on every refresh (`:439`). If container offset 2's entry is missing/transiently absent when `SendIn` resolves it, **replica 2 silently misroutes to node 0**; the true 3rd node never receives the task, so its response never arrives and the op hangs. This fits manifestation (A) with `swim.enabled:false` and all nodes healthy, and explains why 2-node (offsets 0/1 always present) works.

- **(B) Premature `send_map` erase from double-counting `completed_replicas_`.** It is incremented by the `SendIn` dead-skip path (`ipc_run2run.cc:217`) and by `FlushStaleStateForNode` (`:856`), *and* by the genuine `RecvOutAggregate` (`:549`). If a replica is counted by both (node flaps dead → skip/flush, then its in-flight response still arrives), `completed` hits `subtasks_.size()` early, `send_map` is erased, and the genuinely-last response logs "not found." This fits manifestation (B) under SWIM flap.

### Suggested confirmation (the logs offered in the issue)

At `kDebug` on the forwarding node for one broadcast:
1. Does `SendIn` log `target_node_id == 2` for replica 2, or `0`? (resolves (A) via `GetContainerNodeId` `:732` hit vs `:738` fallback.)
2. What is `completed_replicas_` when the "not found in send_map" warning fires, and is there a prior dead-skip (`:217`) / flush-stale (`:856`) increment for that origin? (resolves (B).)

Happy to follow up with the routing/counting fix (e.g. make `GetContainerNodeId` return `kInvalidNodeId` on missing mapping and have `SendIn` treat it as a hard error rather than a silent node-0 reroute) once the logs pin which hazard fires.

### DIFF
--- a/context-runtime/include/clio_runtime/task.h
+++ b/context-runtime/include/clio_runtime/task.h
@@ -41,6 +41,7 @@ using pid_t = int;
 #endif
 
 #include <atomic>
+#include <chrono>
 #ifndef __NVCOMPILER
 #include <coroutine>
 #endif
@@ -627,6 +628,13 @@ struct RunContext {
       0; /**< Predicted wall time from InferWallClockTime */
   TaskStat
       predicted_stat_; /**< TaskStat used for prediction (for reinforcement) */
+  // When this task's replicas were dispatched cross-node via SendIn (steady
+  // clock). ScanSendMapTimeouts uses it as an absolute deadline so an origin
+  // whose replica response never returns — e.g. a response misrouted to the
+  // wrong node because hostfile ordering differs across the cluster (#628) —
+  // still completes with a timeout RC instead of hanging forever. Default
+  // (epoch) means "not cross-node dispatched" and is skipped by the scan.
+  std::chrono::steady_clock::time_point send_enqueue_time_{};
 
   RunContext()
       : coro_handle_(),
@@ -670,7 +678,8 @@ struct RunContext {
         predicted_load_(other.predicted_load_),
         wall_timer_(other.wall_timer_),
         predicted_wall_us_(other.predicted_wall_us_),
-        predicted_stat_(other.predicted_stat_) {
+        predicted_stat_(other.predicted_stat_),
+        send_enqueue_time_(other.send_enqueue_time_) {
 #ifndef __NVCOMPILER
     other.coro_handle_ = nullptr;
 #else
--- a/context-runtime/src/ipc/ipc_run2run.cc
+++ b/context-runtime/src/ipc/ipc_run2run.cc
@@ -176,6 +176,12 @@ void IpcManagerRun2Run::SendIn(ctp::ipc::FullPtr<chi::Task> origin_task) {
   }
   chi::RunContext *origin_rctx = origin_task->GetRunCtx();
 
+  // Stamp the dispatch time so ScanSendMapTimeouts can apply an absolute
+  // deadline: if a replica response never comes back (lost, or misrouted to a
+  // node whose send_map can't match the net_key — #628), the origin still
+  // completes with a timeout RC instead of stalling forever.
+  origin_rctx->send_enqueue_time_ = std::chrono::steady_clock::now();
+
   const std::vector<chi::PoolQuery> &pool_queries = origin_rctx->pool_queries_;
   size_t num_replicas = pool_queries.size();
   origin_rctx->subtasks_.resize(num_replicas);
@@ -775,11 +781,14 @@ void IpcManagerRun2Run::ScanSendMapTimeouts() {
   auto *ipc_manager = CLIO_IPC;
   auto now = std::chrono::steady_clock::now();
 
+  // NOTE: we do NOT early-return when no nodes are dead. The dead-node check
+  // below only catches origins blocked on a node SWIM has marked dead; it
+  // misses the #628 failure where a replica response is misrouted to a *live*
+  // node (hostfile ordering differs across the cluster, so the responder's
+  // ret_node index resolves to the wrong host) and therefore never reaches the
+  // origin's send_map. Those origins would hang forever. The absolute-deadline
+  // pass added below completes them regardless of node liveness.
   const auto &dead_nodes = ipc_manager->GetDeadNodes();
-  if (dead_nodes.empty()) {
-    return;
-  }
-
   std::unordered_map<chi::u64, std::chrono::steady_clock::time_point> dead_map;
   for (const auto &entry : dead_nodes) {
     dead_map[entry.node_id] = entry.detected_at;
@@ -818,6 +827,22 @@ void IpcManagerRun2Run::ScanSendMapTimeouts() {
             }
           }
 
+          // Absolute deadline (#628): even if no target node is dead, an origin
+          // that has waited past its timeout for a replica response that will
+          // never arrive (lost or misrouted to the wrong node) must complete
+          // rather than hang. send_enqueue_time_ == epoch means the task was
+          // never cross-node dispatched, so skip it.
+          if (!any_timed_out &&
+              rctx->send_enqueue_time_ !=
+                  std::chrono::steady_clock::time_point{}) {
+            float waited =
+                std::chrono::duration<float>(now - rctx->send_enqueue_time_)
+                    .count();
+            if (waited >= task_timeout) {
+              any_timed_out = true;
+            }
+          }
+
           if (any_timed_out) {
             to_complete.emplace_back(key, origin_task);
           }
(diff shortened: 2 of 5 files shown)


hyoklee:
Tested this PR in a live 3-node cross-OS cluster — it's a real improvement but a **partial fix**: the 3-node data path still hangs.

**Environment:** 3-node Chimaera/CTE cluster, ZeroMQ/TCP:9413, `swim.enabled: false`. node0 = nene (macOS, x86-64), node1 = jelly (CentOS 7, x86-64), **node2 = raven (Windows 10, x86-64)**. nene + jelly built **with this PR** (`548779e`); Raven is an unmodified rebuild. The 2-node nene+jelly CTE data path works perfectly; adding the Windows 3rd node reproduces #628.

### What this PR fixes (confirmed)
- `RecvOutDeserialize` now downgrades the per-replica miss to a WARNING and `continue`s instead of the old FATAL `return 5`:
  ```
  :476 WARNING RecvOutDeserialize [RecvOut] Task(... replica:2 ... net_key:140569683710672)
        skipped: origin not in send_map — stale/duplicate replica response
  ```
- **No SIGSEGV** — the `EndTask` null-`run_ctx` guard is in place and the earlier crash path (abort → sibling starvation → dead-node recovery → null deref) is gone.

### What still hangs
A CTE op that requires node 2 (Raven) to respond (`PutBlob` via `DirectHash(2)`, i.e. the broadcast replica on node 2) still hangs the client (150 s timeout). The new `[SendOut]`/`[RecvIn]` kInfo breadcrumbs added here pinpoint why:
```
[RecvIn]  ... replica:1 node:0 net_key:140569683710672      (origin = node-0/nene task ptr, pid 72048)
[SendOut] ... replica:1 -> node 0: echoing net_key 140569683710672
[RecvOut] ... replica:2 ... net_key:140569683710672  skipped: origin not in send_map (size 1)   (on jelly)
:525 RecvOutAggregate  Admin: Origin task not found in send_map with net_key 140569683710672
```
The **replica-2 (Raven) response carries `net_key:140569683710672`, which is node-0/nene's task pointer — but it is delivered to jelly**, whose `send_map` only holds jelly's own keys, so it is unmatchable and skipped. `RecvOutAggregate` (:525) already `continue`s as well, so neither RecvOut path aborts now — but *skipping* an unmatchable replica does not *complete* the origin, so `subtasks_[2]` never finishes and the broadcast's origin task never resolves → the op stalls (instead of crashing).

### Takeaway
Necessary (no more crash / batch-starvation) but not sufficient for the heterogeneous 3-node path. The remaining root cause is that the **3rd-node replica response's `net_key` does not resolve to a `send_map` entry on the node that receives it** — it carries the origin node's pointer but is delivered to a different node. A complete fix likely needs either correct origin-node routing for replica responses (so the reply reaches the node whose `send_map` holds that net_key), or a path to drive origin-subtask completion when a replica response is permanently unmatchable (e.g. a per-replica timeout that still completes the aggregate).

Happy to capture Raven's own `[RecvIn]`/`[SendOut]` breadcrumb lines (the responder side) and full 3-node logs if that helps localize where the net_key gets attributed to the wrong node. Context: #628.

lukemartinlogan:
I'm working on integrating these changes into the current dev branch. Writing a test to stress this case and reproducing. The code will probably need to change significantly from what this PR holds.

lukemartinlogan:
Ok, I have added a new docker CI test that specifically stresses cases where replica responses are handled in case of node failures. Branch dev probably could use a re-run on the system that produced this error, but I'm closing this PR.

PR #648, opened 2026-06-30
bench(thallium): raw libthallium async-RPC microbenchmark (na+sm)

### LINKED ISSUE #647: Benchmark: libthallium async RPC throughput/latency over the shm (na+sm) transport
## Motivation

We want to understand how well **libthallium** itself handles **async RPCs**, independent of the lightbeam (`lbm`) transport wrapper. This gives us a clean baseline for the raw RPC engine performance that the runtime's networking sits on top of.

## What to build

A standalone microbenchmark under `context-runtime/benchmarks/thallium-bench` that:

- Uses thallium **directly** (not `ctp::lbm::ThalliumTransport`).
- Uses the **shared-memory transport** (`na+sm`) so we measure intra-node RPC engine overhead without a real NIC.
- Drives **async RPCs** (`rpc.on(ep).async(...)`) so many requests can be in flight at once.

## Configurable parameters

- **RPC threads** — the engine's `rpc_thread_count` (number of Argobots ES running RPC handlers).
- **Batch size** — number of async RPCs submitted in parallel before waiting on the batch.
- (plus total iterations and payload size for sweeps)

## Metrics

- **Overall IOPS** (RPCs/sec).
- **Average latency** per RPC.

## Location

`context-runtime/benchmarks/thallium-bench`, built only when `CLIO_CORE_ENABLE_THALLIUM=ON`.

lukemartinlogan:
## Summary

Adds `context-runtime/benchmarks/thallium-bench`, a standalone microbenchmark that drives **libthallium async RPCs directly** (not through the `ctp::lbm::ThalliumTransport` wrapper) over the **shared-memory transport (`na+sm`)**. It measures how well thallium itself handles async RPCs on an intra-node round trip, with no real NIC involved.

## Design

- Server handler and client submitter share a single `tl::engine` (server mode, dedicated progress thread).
- The client looks up the engine's own self address and fires async RPCs (`rpc.on(ep).async(...)`) at it over the `na+sm` shared-memory loopback.
- The handler is trivial (echoes the payload size) so the measurement reflects RPC-engine overhead, not handler work.

## Configurable

- `--rpc-threads N` — engine `rpc_thread_count` (Argobots handler ES pool).
- `--batch B` — async RPCs kept in flight at once before draining.
- plus `--iters`, `--payload`, `--warmup`, `--protocol`.

## Metrics

- Overall **IOPS** (RPCs/sec).
- **Average latency** per RPC (amortized; use `--batch 1` for isolated round-trip latency).

## Build

Built only when `CLIO_CORE_ENABLE_THALLIUM=ON`; links the `thallium` imported target directly. Default builds (thallium off) are unaffected — the `add_subdirectory` is guarded.

## Verification

Thallium is not installed in this environment, so the source was compile-checked with `-Wall -Wextra` against a minimal mock of the thallium API surface used (engine ctor, `define`/`lookup`/`self`/`finalize`, `on().async()`, `async_response::wait()`). The engine ctor and define/handler/lookup signatures mirror the existing `ThalliumTransport` in `context-transport-primitives`.

Closes #647

### DIFF
--- a/context-runtime/CMakeLists.txt
+++ b/context-runtime/CMakeLists.txt
@@ -80,6 +80,13 @@ add_subdirectory(modules)
 if(CLIO_CORE_ENABLE_BENCHMARKS)
   add_subdirectory(benchmark)
   message(STATUS "Benchmarks enabled - added benchmark subdirectory")
+
+  # Raw libthallium async-RPC microbenchmark (issue #647). Only builds when
+  # thallium is available; it links the `thallium` imported target directly.
+  if(CLIO_CORE_ENABLE_THALLIUM)
+    add_subdirectory(benchmarks/thallium-bench)
+    message(STATUS "Thallium enabled - added benchmarks/thallium-bench")
+  endif()
 endif()
 
 # Add test subdirectory if testing is enabled
--- a/context-runtime/benchmarks/thallium-bench/CMakeLists.txt
+++ b/context-runtime/benchmarks/thallium-bench/CMakeLists.txt
@@ -1,0 +1,13 @@
+# thallium-bench: raw libthallium async-RPC microbenchmark (issue #647).
+#
+# Built only when CLIO_CORE_ENABLE_THALLIUM=ON. Uses thallium directly (not the
+# ctp::lbm transport) over the na+sm shared-memory transport to measure async
+# RPC IOPS and latency.
+
+add_executable(thallium_bench thallium_bench.cc)
+
+# `thallium` is the imported target provided by find_package(thallium) in the
+# root CMakeLists.txt (guarded by CLIO_CORE_ENABLE_THALLIUM).
+target_link_libraries(thallium_bench PRIVATE thallium)
+
+install(TARGETS thallium_bench DESTINATION bin)
--- a/context-runtime/benchmarks/thallium-bench/README.md
+++ b/context-runtime/benchmarks/thallium-bench/README.md
@@ -1,0 +1,68 @@
+# thallium-bench
+
+Raw **libthallium** async-RPC microbenchmark (issue #647).
+
+This measures how well thallium itself drives async RPCs, independent of the
+lightbeam (`ctp::lbm`) transport wrapper. It talks to thallium **directly** over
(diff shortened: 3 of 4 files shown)


lukemartinlogan:
## Built and run end-to-end ✅

Built against a real **thallium 0.11.3 / mercury 2.3.1 / mochi-margo 0.17.0 / argobots 1.2** stack (spack) and run over the **`na+sm`** shared-memory transport. Clean exit (code 0) after fixing the engine shutdown ordering (see follow-up commit).

Single-node box, single submitter thread, empty payload, 100k timed RPCs each.

### Batch sweep (`--rpc-threads 4`)

| batch | IOPS | avg latency |
|------:|-----:|------------:|
| 1   | 59,054  | 16.93 µs |
| 2   | 95,381  | 10.48 µs |
| 4   | 129,820 |  7.70 µs |
| 8   | 155,224 |  6.44 µs |
| 16  | 156,869 |  6.37 µs |
| 32  | **174,014** | **5.75 µs** |
| 64  | 159,368 |  6.27 µs |
| 128 | 134,163 |  7.45 µs |
| 256 | 115,561 |  8.65 µs |

Async batching is clearly worth it: throughput ~3× from batch 1→32, then over-subscription past ~32 starts hurting.

### Thread sweep (`--batch 32`)

| rpc-threads | IOPS | avg latency |
|------------:|-----:|------------:|
| 1  | **289,856** | **3.45 µs** |
| 2  | 230,205 | 4.34 µs |
| 4  | 174,153 | 5.74 µs |
| 8  | 160,327 | 6.24 µs |
| 16 | 163,708 | 6.11 µs |

Interesting finding: for a **single submitter doing tiny RPCs over shared memory, more handler threads *reduce* throughput** — 1 ES peaks at ~290K IOPS / 3.45 µs. With trivial per-RPC work the Argobots cross-ES scheduling/locking overhead dominates, so the async pipeline (batch), not the handler pool, is what drives concurrency here. Multi-submitter / larger-payload sweeps would shift this; the knobs are there to explore it.

PR #650, opened 2026-06-30
feat(ml): integrate predictive ML drive failure modeling

yuvankas1:
Integrates the ML drive failure prediction model with the C++ backend. Includes static prediction models and rolling history capture.


lukemartinlogan:
I've updated on my branch to use git LFS instead for the large files. I'll close this PR soon.

Issue #672, opened 2026-07-02
Performance updates to BDEV Module


lukemartinlogan:
# Integrate GPU I/O Libs into CLIO directly

Let's copy-paste the following repository under context-transfer-engine/adapters/gpu-hdf5: https://github.com/lukemartinlogan/gpu-io-libs

I will delete that repository afterwards. This way, the code will evolve with the CLIO platform more naturally to avoid conflicts in the future.

We should add a new CLIO_ENABLE_GPU_HDF5 flag to cmake to gate this. It should require CLIO_CORE_ENABLE_CUDA=ON as well to compile. 

Ensure that all the code compiles with the new APIs of this repo. The code in this copy-paste is using deprecated APIs (e..g, chi:: and CHIMAERA_*)

# Upgrade the BDEV Module to Support Asynchronous Memcpy

Update the memory bdev transport to support Asynchronous memcopy from gpus. 
Right now all memory copies are fully synchronous, which is bad for performance.
We can have separate functions for handling GPU for CPU memory copy paths within the mem bdev.

PR #679, opened 2026-07-03
feat: Implement async operations for bdev module

abscosmos:
This also adds tests and a benchmark. Related: #672.


abscosmos:
@lukemartinlogan I'm pretty sure the failing test is a flake, since it's a file lock race and isn't connected to any files this PR touched.

PR #690, opened 2026-07-07
CFS: route exact-path tag/blob ops by path hash for cross-node visibility (#685)

### LINKED ISSUE #685: Multi-node: CTE filesystem (CFS) tags are node-local (PoolQuery::Local) — shared files invisible across nodes
## Summary

In a multi-node runtime, the CTE **filesystem (CFS)** adapter's tag namespace is **node-local**: a file created through the POSIX interceptor (`libclio_cte_posix.so`) on one node is not visible to the interceptor on any other node. Every path-keyed tag operation in `context-transfer-engine/filesystem/src/filesystem_runtime.cc` is issued with `clio::run::PoolQuery::Local()`, which routes only to the local node's CFS container. As a result the "shared filesystem" is effectively per-node, and cross-node shared-file access fails.

## Environment

- clio-core `dev` @ `f9e7273d`
- TACC Vista, 2 × Grace-Hopper/Grace nodes (ARM aarch64), gcc-15 (conda), ELF=ON build
- 2-node runtime bootstrapped over InfiniBand (ib0): one daemon/node, ZeroMQ bound to each node's ib0 IP, `neighborhood=2`; both daemons compose all pools and connect bidirectionally (verified). Same-node `clio::` round-trips pass on each node.

## Reproduction

1. Start a 2-node CTE runtime (`clio_cte_core` @ 512.0 + `clio_cte_filesystem` @ 560.0, `targets.neighborhood: 2`), one daemon per node.
2. On **node A**, via `LD_PRELOAD=libclio_cte_posix.so`, `open()`+`write()`+`close()` a file named `clio::/some/path` (rc=0; same-node read-back on A succeeds).
3. On **node B**, via the same `LD_PRELOAD`, `open()` the same `clio::/some/path`.

**Expected:** node B opens the file created on node A (shared namespace across the cluster).
**Actual:** `open()` on node B fails with **ENOENT** — the tag created on A doesn't exist in B's local CFS container.

## Root cause

`filesystem_runtime.cc` uses `PoolQuery::Local()` for *all* tag operations — `AsyncGetOrCreateTag`, `AsyncTagQuery`, `AsyncGetTagSize`, `AsyncDelTag`, append-staging, etc. (~30 call sites, e.g. lines 106, 142, 159, 171, 268, 305, 349, 455, 460, 500…). `PoolQuery` in `context-runtime/include/clio_runtime/pool_query.h` documents `Local` as "Route to local node only", so a path's tag is created/looked-up only in the issuing node's container. There is no consistent cross-node ownership for a given path, so the same path resolves to different (or missing) tags on different nodes.

## Suggested fix

Route **exact-path** tag operations by a deterministic hash of the path so a given path is always owned by the same container cluster-wide, instead of the local node:

- Replace `PoolQuery::Local()` with `PoolQuery::DirectHash(hash(path))` for the single-path ops (`Open`/`GetOrCreateTag`/`GetTagSize`/`DelTag`/read/write/append on a specific path). `DirectHash` ("Route using hash-based load balancing") already exists in `PoolQuery` and gives consistent path→container routing from any node.
- **Directory / prefix scans** (`AsyncTagQuery` with a child regex used by `readdir`/`getattr`) must instead fan out across the cluster — `PoolQuery::Broadcast()` — since matching children may live in any container; results then merged.
- The `_clio_append_staging` flat tag and the append-sequence logic need a fixed, cluster-consistent owner (e.g. `DirectHash` of a constant key) so appends to a file converge on one container.

Happy to submit a PR (see linked). Note: care is needed to avoid the multi-node routing deadlocks that have been observed under the distributed runtime; the PR scopes the change to the single-file data path first and treats directory-scan fan-out as a follow-up.

hyoklee:
Fixes #685.

## Problem

The CTE filesystem (CFS) tag namespace is node-local: every path-keyed op in `context-transfer-engine/filesystem/src/filesystem_runtime.cc` uses `PoolQuery::Local()`, so a file created through the POSIX interceptor on one node is invisible on any other node (`open()` → ENOENT). See #685 for the full reproduction.

## Change

Route **exact-single-path** tag/blob ops by a deterministic hash of the path so a given path is owned by — and re-resolved to — the same container from any node. New helper:

```cpp
inline clio::run::PoolQuery PathQuery(const std::string &path) {
  clio::run::u32 hash_value =
      static_cast<clio::run::u32>(std::hash<std::string>{}(path));
  return clio::run::PoolQuery::DirectHash(hash_value);
}
```

This intentionally mirrors the CTE **core's own** tag routing in `core_runtime.cc` (`GetOrCreateTag`: `static_cast<u32>(std::hash<std::string>(tag_name))` → `DirectHash`), so the CFS layer and the core agree on which container owns a path's tag.

`Local()` → `PathQuery(path)` for the exact-path ops only: **Open** (TagQuery / GetOrCreateTag / GetTagSize), **Read** (GetBlob), **Write** (PutBlob), **Getattr** (file branch), **StatSize**, **Truncate**, **Unlink** (DelTag). One file changed (+41/−17).

## Scope (deliberately minimal)

Left as `Local()` and called out as follow-up work: the directory/prefix `child_re` scans (`readdir`/`getattr`/`mkdir`/`rmdir`/`rename`/`link`) and the deferred-append pipeline + `_clio_append_staging` tag. Those need cluster-wide fan-out (e.g. `PoolQuery::Broadcast()` with result merge) and are a larger change; this PR targets single-file cross-node open/read/write first.

## Validation status — DRAFT

- **Builds cleanly** on TACC Vista (ARM aarch64, conda gcc-15, ELF=ON); `libclio_cte_posix.so` and `libclio_cte_filesystem_runtime.so` both cap at `GLIBC_2.34` (load-compatible).
- **Cross-node read-back is NOT yet validated end-to-end**, blocked by a *separate, pre-existing* crash unrelated to this patch: on `dev@f9e7273d` the POSIX interceptor **client** SEGVs in `ctp::lbm::Transport::Send<SaveTaskArchive>` (via `IpcCpu2CpuZmq::SendIn<filesystem::OpenTask>` ← `CfsIo::Open` ← `open()`) when submitting the **first** filesystem task. It reproduces on a **single node** in **both SHM and TCP IPC modes** — i.e. before any cross-node routing and before this PR's CFS code runs (`nm -DC` confirms the patched `Runtime::Open` is only in the daemon-side `libclio_cte_filesystem_runtime.so`, absent from the crashing client `libclio_cte_posix.so`). The 2-node cluster itself is healthy (ROUTERs bound over InfiniBand, peers connected, all pools composed).

Marking this **draft** until that client-side transport SEGV on ARM is triaged (or the change is validated on a platform where the interceptor can submit tasks). The fix itself is minimal and consistent-by-construction with the core's routing; review of the approach (and guidance on the readdir/append fan-out) is welcome.

Tested-on: TACC Vista, 2× NVIDIA Grace (ARM aarch64), InfiniBand ib0, OpenMPI/ZeroMQ.

### DIFF
--- a/context-transfer-engine/filesystem/src/filesystem_runtime.cc
+++ b/context-transfer-engine/filesystem/src/filesystem_runtime.cc
@@ -62,6 +62,30 @@ inline std::string StripTrailingSlash(std::string p) {
   if (p.size() > 1 && p.back() == '/') p.pop_back();
   return p;
 }
+/**
+ * Route an exact-single-path tag/blob op to the container that deterministically
+ * owns `path` cluster-wide, instead of PoolQuery::Local() (which only ever hits
+ * the caller's own node). Without this, a file's tag+page-blobs are created on
+ * whichever node the writing process ran on, so a file written through the POSIX
+ * interceptor on node A is invisible on node B (open -> ENOENT). (#685)
+ *
+ * The hash mirrors the CTE core's OWN GetOrCreateTag scheduling
+ * (core_runtime.cc: static_cast<u32>(std::hash<std::string>(tag_name)) ->
+ * DirectHash), so the container a path's tag is created on is exactly the one
+ * every node re-derives for the follow-up TagQuery / GetTagSize / Put/GetBlob /
+ * DelTag on that same path — keeping a single file's whole data path colocated
+ * on one container and consistent from any node.
+ *
+ * NOTE (scope, #685): only EXACT-single-path ops route this way. Directory /
+ * prefix regex scans (readdir/getattr child_re) stay Local() for now — a
+ * documented follow-up — since a per-node tag namespace makes a global prefix
+ * scan a broadcast+merge problem, out of scope for this single-file-path fix.
+ */
+inline clio::run::PoolQuery PathQuery(const std::string &path) {
+  clio::run::u32 hash_value =
+      static_cast<clio::run::u32>(std::hash<std::string>{}(path));
+  return clio::run::PoolQuery::DirectHash(hash_value);
+}
 /** Escape regex metacharacters for an exact TagQuery match (from libfuse). */
 inline std::string EscapeExact(const std::string &s) {
   std::string out;
@@ -139,7 +163,7 @@ clio::run::TaskResume Runtime::Open(clio::run::shared_ptr<OpenTask> &task) {
   // Did the tag already exist (so we can report created_)?
   bool existed = false;
   {
-    auto q = cte_.AsyncTagQuery(EscapeExact(path), 1, clio::run::PoolQuery::Local());
+    auto q = cte_.AsyncTagQuery(EscapeExact(path), 1, PathQuery(path));
     CLIO_CO_AWAIT(q);
     existed = (q->GetReturnCode() == 0 && !q->results_.empty());
   }
@@ -157,7 +181,7 @@ clio::run::TaskResume Runtime::Open(clio::run::shared_ptr<OpenTask> &task) {
 
   // Resolve / create the tag for this path.
   auto t = cte_.AsyncGetOrCreateTag(path, clio::cte::core::TagId::GetNull(),
-                                    clio::run::PoolQuery::Local());
+                                    PathQuery(path));
   CLIO_CO_AWAIT(t);
   if (t->GetReturnCode() != 0) {
     task->return_code_ = EIO;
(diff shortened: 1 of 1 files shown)


hyoklee:
Cross-node validation of this fix is currently blocked by #697 — the POSIX interceptor client SIGSEGVs in `ctp::lbm::Transport::Send` on the first filesystem `OpenTask` (single-node, both SHM and TCP IPC), so no task can be submitted from the interceptor to exercise the read-back. Keeping this PR in **draft** until #697 is resolved (or the change is validated on a platform where the interceptor can submit tasks). The change here is confined to the daemon-side CFS routing and does not touch the crashing client transport path.

lukemartinlogan:
You should pull from dev main before continuing this work. It may be fixed.

hyoklee:
Superseded by `f58c7ca7` ("fix(cte): route CFS ops via Dynamic") on `dev`, which resolves #685 by routing CFS ops through `PoolQuery::Dynamic()` — a cleaner approach than the per-path `DirectHash` in this PR — and lands an official cross-node CFS test (#689). Closing in favour of that. Thanks!

PR #699, opened 2026-07-07
Harden clio-core HDF5 VOL with test suite and telemetry

mattjala:
This PR is a general hardening of the clio-core `iowarp` HDF5 VOL connector.

### 1. Compatibility suite/Regression gate
A test suite that checks that the VOL preserves native HDF5 semantics. For each feature case
it compares I/O done through the connector against the same I/O done with native HDF5.

It routes writes and reads four ways (native<->native as the reference, VOL-write->native->read,
native-write->VOL-read, and a full VOL round-trip) and also runs native `h5diff`/`h5dump` on the results. The
round-trip arm serves reads back from the CTE cache, so a bad cache is caught as
wrong data. Wired into CI as `cte_hdf5_vol_compat_suite`.

### 2. Compatibility fixes
- **Variable-length datatypes (strings + attributes)** — root cause was a missing
  file post-open handshake: the connector reported no optional-op support and left
  `file_optional` null, so `H5F_VOL_OBJ` stayed NULL and vlen create dereferenced
  NULL. Fixed by forwarding `introspect_opt_query` and `file_optional` to the
  under-VOL (mirrors `H5VLpassthru`).
- **Groups / links** — unwrap the hard-link target in `link_create`.
- **Iteration** — not a VOL bug: h5py calls deprecated APIs HDF5 restricts to the
  native VOL. The h5py case is disabled and replaced by an isolated C test using
  the modern `H5Ovisit3`/`H5Literate2` (`c_iteration`).

### 3. Safer file flushes/closes
`H5Fflush`/`H5Fclose` now drain every in-flight async CTE put before returning.
Tested by `c_safeflush`.

### 4. Inital work on serve-only selection-aware read caching (serve-only)
Hyperslab/point reads can now be served from the CTE cache instead of falling back
to native. Since the cache stores each dataset as one linear byte image (not indexed
by selection), the connector materializes the whole image from the CTE blobs into
memory, then uses HDF5's own gather/scatter to extract the requested selection into
the user's buffer.

On a cache miss it falls back to native unchanged.

The current version fetches the full image per selection, and so no optimization is expected.
Range-limited fetched of only the touched chunks is future work.

This unearthed a problem with key invalidation on a non-whole write that was fixed.

It also revealed a complexity with compound/array datatypes (and likely other) datatypes, since the in-memory buffer can diverge from the on-disk image due to padding and mem/file type conversion. This will need some design decisions made eventually, but for now I've limited it to just atomic datatypes where this isn't a problem.

Collective MPI I/O is also out of scope for now.

### 5. Access telemetry / observability (observe-only)
Enabled by the `IOWARP_VOL_TRACE` environment variable, which should be populated with an output directory.

When enabled, for each open HDF5 file the VOL will write two files into that directory, named `<hdf5-basename>.<pid>.
access.{jsonl}` and `<hdf5-basename>.<pid>.access.{json}`. The PID is included to prevent concurrent process/ranks from clobbering each other.

The JSONL file is live per-access, with one record appened per dataset read/write.
The JSON file is written to once at file close and serves as an aggregated summary JSON.

The summary is intended to carry actionable signals for agents to use in the future: read cache hit rate, repeated selections, transfer-size distribution, chunk alignment (rechunk signal), and cache-vs-native latency.

### DIFF
--- a/.github/workflows/ci-adapters.yml
+++ b/.github/workflows/ci-adapters.yml
@@ -61,7 +61,15 @@ jobs:
               # FUSE backend + ELF interception headers (libelf/HDF5 already in image)
               apt-get update -qq
               apt-get install -y -qq libfuse3-dev fuse3 pkg-config libelf-dev
+              # h5py for the VOL compat suite, SOURCE-built (--no-binary) against
+              # the image HDF5 (/usr/local, 2.1.1) so the connector loads into
+              # h5py. A PyPI wheel bundles its own libhdf5 and cannot load a VOL
+              # built against the system HDF5. Pin the suite to this interpreter
+              # via -DPython3_EXECUTABLE below.
+              HDF5_DIR=/usr/local /usr/bin/python3 -m pip install \
+                --break-system-packages --no-binary=h5py h5py numpy
               cmake -B build-adapters \
+                -DPython3_EXECUTABLE=/usr/bin/python3 \
                 -DCMAKE_BUILD_TYPE=Debug \
                 -DCLIO_CORE_ENABLE_RUNTIME=ON \
                 -DCLIO_CORE_ENABLE_CTE=ON \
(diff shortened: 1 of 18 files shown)


lukemartinlogan (changes requested):
Minor fix: could you rename iowarp_vol -> clio_vol

Also, I recommend using an agent loop to poll the CI to check for failures and fix them if they are coming from PRs. I have found that tests often pass on local machines and then fail on actions in some way shape or form

Also, please rebase this to go to dev instead of main

lukemartinlogan (approved):

Issue #526, opened 2026-06-11
Reproducible jarvis pipelines for performance evaluations


lukemartinlogan:
We would like to have the ability to rerun iowarp experiments daily in Ares to ensure that our updates don't cause performance regressions. Jarvis is a perfect system for this.

In the new jarvis readme, the example provided demonstrates how to build a configuration with multiple experiments attached.

We also have updated doc under docs/pipeline_tests.md that describes the advanced full specification of jarvis.

Here is an example of a single pipeline that uses CTE+libfuse: jarvis_clio_core/pipelines/ares/ior/ior_cte_libfuse_sbatch.yaml

Create a new subdirectory jarvis_clio_core/pipelines/ares/CI. You should create two pipeline tests: single_node.yaml and distributed.yaml. These files should be fully self-contained. Meaning they should include the installation + slurm batch deployment of the software.

# Single-node Experiment

Let's have the following parameter sweeps:
1. Vary I/O size between 4KB, 64KB, 128KB, and 1MB
2. Vary  thread counts between 1, 2, 4, 8, 16, 32

Run the following experiments in the pipeline:
1. NFS + ior
2. CTE + libfuse + ior
3. redis + redis-bench
4. juicefs

Ideally, the pipelines would install juicefs, redis, and CTE. 
You can use containerized deployments for juicefs and redis. CTE you should use ``spack install iowarp +libfuse``.

# Multi-Node Experiment

1. Vary the number of nodes between 1, 2, and 4

Run the following experiments in the pipeline:
1. ior + NFS
2. CTE + libfuse + ior
3. juicefs

PR #717, opened 2026-07-09
feat(#526): reproducible Apptainer-containerized Jarvis regression pipelines

eDoggo3779:
## Summary

Two daily-regression pipelines that run **every** benchmark/service command inside a running Apptainer instance, so all binaries come from the SIF with **zero host-env dependency** beyond jarvis + apptainer + the SIF itself. Based on `dev`.

### Pipelines

| Pipeline | Coverage | Status |
|---|---|---|
| `pipelines/ares/single_node.yaml` | 4 storage stacks (NFS fio, CTE+libfuse fio, redis via `clio_redis_bench`, JuiceFS fio) × 4 io_size × 6 threads = 24 combos → one `results.csv` | 24/24 green on a clean run; see [Known issue](#known-issue-stale-runtime-state-and-a-hard-kill-teardown-still-open-on-the-clio-side) |
| `pipelines/ares/distributed.yaml` | Multi-node CTE (`clio_cte_bench`) over `nprocs [1,2,4]` in ONE 4-node allocation | 3/3 ✅ |

### Container model

`base_deploy_mode: container` + one apptainer instance per sweep run. Every package threads `container`/`container_image` `ExecInfo` kwargs (`container_utils.container_kwargs`) so `ExecFactory` wraps each command as `apptainer exec instance://<run> ...`.

- FUSE mounts (CTE libfuse, JuiceFS) need `--fakeroot` on rootless apptainer.
- `/tmp` is bind-rooted to node-local disk (`tmp_bind_root`) for runtime IPC + JuiceFS cache.

### What's in this PR

- **New packages:** `juicefs`, `juicefs_bench`, `container_utils`
- **Containerized + hardened:** `redis`, `clio_runtime`, `clio_cte`, `clio_cte_libfuse`, `clio_cte_bench`, `clio_redis_bench`
  - multi-host metric folding
  - wrapped `redis-cli` helper
  - timeout-bounded runtime stop under **both** pre/post-rebrand daemon names
  - pre-start stale-daemon reap
- **Ops:** sbatch wrappers (`run_single_node_ares`, `run_distributed_ares`) with:
  - per-node stale-state cleanup (ghost runtimes holding port 9413, leaked `/dev/shm/chi_*`, stale `/tmp` scratch)
  - an ssh preflight that starts a **real** `--fakeroot` instance per node and, on failure, names the bad node(s) + prints a ready `sbatch --exclude=<csv>` resubmit
  - hostfile pinning
  - `validate_apptainer.sh` preflight checklist
  - SIF build chain (`Dockerfile.regression` + Docker-free `regression.def` + `build_regression_image.sh`), gated on every required binary being present

- **Spack:** `iowarp` recipe adds a `+redis` variant (builds upstream's `clio_redis_bench`; `hiredis` dep). The regression image builds `iowarp@dev +fuse +redis` — upstream `dev` source with the redis bench enabled via this recipe overlay; **no fork pin**. Once this PR merges, `+redis` is upstream and the overlay is unnecessary.

**Scope:** this is the tight core — the two regression pipelines plus exactly the packages/scripts they need. Adjacent experiments (standalone JuiceFS-fio and JuiceFS-over-GCS pipelines + their setup/smoke scripts, and the CTE 1-node sweep) are intentionally omitted.

---

## Running it on Ares

**One-time setup (login node).** After this, every rerun is a single `sbatch` line. Assumes the standard Ares baseline: a miniconda `iowarp` env with jarvis-cd `dev @ 2925ee8` already `jarvis init`-ed (this is what supplies `shared_dir` under `/mnt/common`).

1. **Get the code** where the wrappers expect it — they default `REPO=$HOME/clio-core-fork`, conda env `iowarp`:
   ```bash
   git clone -b pipelines-ship https://github.com/eDoggo3779/clio-core-fork.git ~/clio-core-fork
   conda activate iowarp && jarvis repo add ~/clio-core-fork/jarvis_clio_core
   ```
2. **Install apptainer + wire it onto the remote PATH** (jarvis Pssh's `apptainer instance start` to every node over ssh, so its spack bin must be on the *non-interactive* shell PATH). This is Phase 0 of the preflight; `APPEND_BASHRC=1` appends the bin to the shared `~/.bashrc`:
   ```bash
   cd ~/clio-core-fork && APPEND_BASHRC=1 jarvis_clio_core/scripts/validate_apptainer.sh
   ```
   (The SIF-dependent gates here are expected-red until step 3.)
3. **Build the SIF once** (`docker build` → `apptainer build`; writes `<shared_dir>/containers/iowarp-regression-526.sif` on `/mnt/common`, so it's visible on every compute node — no per-node copy):
   ```bash
   cd ~/clio-core-fork && jarvis_clio_core/scripts/build_regression_image.sh
   ```
4. **Green preflight** — reruns Gates 0–3 (apptainer on login+compute, jarvis pin, SIF present, a real FUSE mount inside the SIF):
   ```bash
   jarvis_clio_core/scripts/validate_apptainer.sh
   ```

**Then run either sweep, as often as you like.** The SIF and PATH wiring persist, so from here it's one line — exclude the known-bad fakeroot nodes on every submit:
```bash
sbatch --exclude=ares-comp-15,ares-comp-17 jarvis_clio_core/scripts/run_single_node_ares.sbatch
sbatch --exclude=ares-comp-15,ares-comp-17 jarvis_clio_core/scripts/run_distributed_ares.sbatch
```
Results land in `~/single_node_results/results.csv` (24 rows) and `~/distributed_results/results.csv` (3 rows). The wrappers default to `--partition=debug` (idle nodes `ares-comp-03..06`); `--exclude` only bites when you override to `--partition=compute` and is a harmless no-op otherwise, so keep it on every submit. Rebuild the SIF (step 3) to track a newer IOWarp; everything else is set-and-forget.

---

## Validation (Ares)

| Pipeline | Result | Evidence |
|---|---|---|
| `distributed.yaml` | **3/3 success** | non-zero `cte_bench.*.agg_bw_mbps` for every combo — Ares job `21564` |
| `single_node.yaml` | **24/24 green** on a clean run (job `21061`, ares-comp-27); a later run (`21563`) hit the CLIO-side hang below | see [Known issue](#known-issue-stale-runtime-state-and-a-hard-kill-teardown-still-open-on-the-clio-side) |

**Bad nodes:** `ares-comp-15` and `ares-comp-17` fail rootless `apptainer instance start --fakeroot` (comp-15 lacks the `uidmap` setuid helpers `newuidmap`/`newgidmap`) → the node is SWIM-marked DEAD and distributed hangs on retry-to-dead-node timeouts. Exclude them: `sbatch --exclude=ares-comp-15,ares-comp-17`. The sbatch preflight also auto-detects this per job and prints the ready resubmit.

---

## Known issue: Runtime C++ bug (not fixed in this PR)

After ~18 sequential runtime start/stop cycles in one `single_node` allocation, a task routed to the null pool:

```
worker.cc: Container not found for pool_id=PoolId(0,0)
```

...and hung indefinitely — no IPC failover, no timeout, so the hang ate the rest of the job's wall clock instead of failing fast.

Root cause traced to the runtime's teardown always **hard-killing** the daemon (`Kill('chimaera')` / SIGKILL) rather than relying on a clean graceful exit, which can leave an escaped process with inconsistent pool-registry state for the next combo to collide with.

> This is CLIO-core (`context-runtime`), out of scope for a jarvis-package PR — confirmed with the CLIO maintainer, who is addressing it upstream.

`jarvis_clio_core/clio_runtime/pkg.py` now force-kills **both** the pre-rebrand (`chimaera`) and post-rebrand (`clio_run`) daemon names as a stopgap, and reaps any stale survivor before each combo's runtime start — but the underlying "hard-kill teardown can strand state" issue is a CLIO fix, not a jarvis one.

---

## Test plan

- [x] `distributed.yaml`: 3/3 combos succeed with non-zero `cte_bench.*` metrics (Ares job `21564`)
- [x] `single_node.yaml`: full 24-combo grid completes green on a clean run (job `21061`); a later run surfaced the CLIO-side hang above, not a pipeline defect
- [ ] Re-validate `single_node.yaml` against a CLIO build once the upstream null-pool/hard-kill fix lands

### DIFF
--- a/installers/spack/packages/iowarp/package.py
+++ b/installers/spack/packages/iowarp/package.py
@@ -46,6 +46,9 @@ class Iowarp(CMakePackage):
     variant('rocm', default=False, description='Enable ROCm support')
     variant('adios2', default=False, description='Build with ADIOS2 support')
     variant('fuse', default=False, description='Enable FUSE3 adapter (CTE)')
+    variant('redis', default=False,
+            description='Build the Redis comparison benchmark '
+                        '(clio_redis_bench); requires hiredis')
     variant('s3', default=False,
             description='Enable Amazon S3 (s3://) import in CAE (aws-sdk-cpp)')
     variant('gcs', default=False,
@@ -80,6 +83,7 @@ class Iowarp(CMakePackage):
     depends_on('hdf5', when='+hdf5')
     depends_on('adios2', when='+adios2')
     depends_on('libfuse@3:', when='+fuse')
+    depends_on('hiredis', when='+redis')
     depends_on('boost+context', when='+boost_coro')
 
     # CAE cloud import backends (opt-in). aws-sdk-cpp for the S3 assimilator
@@ -233,6 +237,10 @@ class Iowarp(CMakePackage):
                 args.append(self.define('CTE_ENABLE_ROCM', 'ON'))
             if '+fuse' in self.spec:
                 args.append(self.define('CLIO_CTE_ENABLE_FUSE_ADAPTER', 'ON'))
+            if '+redis' in self.spec:
+                # Builds clio_redis_bench (the apples-to-apples redis peer of
+                # clio_cte_bench) — the #526 single_node redis experiment.
+                args.append(self.define('CLIO_CORE_ENABLE_REDIS', 'ON'))
 
         # Context-assimilation-engine (CAE) options (if enabled)
         if '+cae' in self.spec:
--- a/jarvis_clio_core/jarvis_clio_core/clio_cte/pkg.py
+++ b/jarvis_clio_core/jarvis_clio_core/clio_cte/pkg.py
@@ -8,6 +8,7 @@ from jarvis_cd.core.pkg import Service
 from jarvis_cd.util import SizeType
 from jarvis_cd.shell.process import Rm, Mkdir
 from jarvis_cd.shell import PsshExecInfo, Exec
+from jarvis_clio_core.container_utils import container_kwargs
 import yaml
 import os
 import re
@@ -166,6 +167,12 @@ class ClioCte(Service):
             parent_dir = os.path.dirname(path)
             if not parent_dir:
                 continue
+            # NOTE: configure runs BEFORE the pipeline's container instance
+            # starts, so this Mkdir cannot be container-wrapped. It is fine
+            # for host-visible paths ($HOME, shared storage); a file device
+            # under /tmp (remapped in-container via tmp_bind_root) would need
+            # its dir created at start() time instead. Both #526 pipelines
+            # use ram:: tiers only, so this loop is a no-op there.
             Mkdir(parent_dir,
                   PsshExecInfo(env=self.mod_env, hostfile=self.hostfile)).run()
 
@@ -226,10 +233,9 @@ class ClioCte(Service):
         Exec(cmd, PsshExecInfo(
             env=self.mod_env,
             hostfile=self.hostfile,
-            container=getattr(self, '_container_engine', 'none'),
-            container_image=self.deploy_image_name(),
             private_dir=self.private_dir,
             bind_mounts=self.container_mounts,
+            **container_kwargs(self),
         )).run()
 
         self.log("CTE started successfully")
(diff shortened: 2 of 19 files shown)


lukemartinlogan (changes requested):
1. juicefs_bench seems to be just fio, which is already apart of jarvis-cd/builtin. We should update the builtin one instead with any differences you needed here.
2. redis is already apart of jarvis-cd/builtin. We shouldn't duplicate.

Overall, let's try consolidating those things back into core jarvis rather than shipping these particular things with CTE. 

Also, please use the scheduler block. I strongly suspect your jarvis installation is outdated and your AI did something unnecssary. We should remove all the "scripts" you made and consolidate them to jarvis

PR #718, opened 2026-07-09
feat(vfd): A native write-through clio-core HDF5 VFD

mattjala:
This PR adds a native write-through mode to the CLIO HDF5 VFD. Before, the driver stashed every byte in CLIO's CTE as page-blobs and nothing ever landed on disk as an actual HDF5 file, so h5dump and similar tools couldn't access it. Now the driver writes straight through to a real on-disk native HDF5 file at the plain path and fsyncs on close, so it's a normal HDF5 file that standard tools read live and matches sec2 byte-for-byte. It opens a CTE handle alongside mirrors writes into it as groundwork for a future read/tiering cache, but reads currently come from the authoritative file.

On top of the core change there's a test suite (multiple dtypes, a multi-page dataset, 2-D / chunked / compressed datasets, groups, attributes, a hyperslab overwrite, two files open at once, plus a differential check against sec2 and an h5dump/h5ls/h5repack/h5diff pass on the file with no VFD loaded), and a dedicated lightweight CI job that builds just the VFD and runs that suite in a few minutes instead (since the full-adapter matrix takes much longer). There's also a hardening pass fixing a few latent bugs (an OOM NULL-deref, read errors getting masked as zeros, some NULL/invalid-fd guards) and a fair bit of dead-code and stale-comment cleanup.

The remaining VFD callbacks (flush/truncate/lock/get_handle), feature-flag advertisement, a real driver FAPL, error-stack wiring, and the actual CTE read cache are all going to be handled in separate follow-ups.

### DIFF
--- a/.github/workflows/ci-vfd.yml
+++ b/.github/workflows/ci-vfd.yml
@@ -53,12 +53,19 @@ jobs:
               # ELF interception headers for the CTE adapters (HDF5 already in image).
               apt-get update -qq
               apt-get install -y -qq libelf-dev pkg-config
-              # Assert the native HDF5 CLI tools are present, so the in-test
-              # native-compat matrix (h5dump/h5ls/h5repack/h5diff on the VFD's
-              # file) actually runs here rather than taking its skip path. set -e
-              # makes a missing tool fail the job instead of silently downgrading.
-              echo "== native HDF5 tools (required for the in-test tool matrix) =="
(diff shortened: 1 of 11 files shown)


lukemartinlogan:
Let's remove the old hermes nomenclature and call clio. The output should be clio_vfd and all macros should say CLIO instead of HERMES

lukemartinlogan (approved):

PR #761, opened 2026-07-17
feat(vfd): feature complete the CLIO HDF5 VFD

mattjala:
This builds on the native write-through core and fills in the rest of what makes the CLIO VFD behave like a real, first-class HDF5 file driver (rather than a minimal byte shim). Four main things:

- Feature flags. query() now advertises the same I/O-optimization flags sec2 does (metadata aggregation/accumulation, data sieve, SWMR). Practical effect: HDF5 lays a file out the same way it would for the default driver, so the VFD's on-disk file is byte-size-identical to sec2 for the same content (and SWMR readers work).
- A real driver FAPL + H5Pset_fapl_clio. You can now select the driver the idiomatic HDF5 way and carry config with it. The one knob today is cache_enabled, which makes CTE cache population opt-in per file — so you can run pure native (sec2-equivalent, no CTE write traffic) or opt into the cache tier.
- Vectored I/O. read_vector/write_vector are implemented (with HDF5's size-array shortening convention), so HDF5's batched and selection-I/O paths hit one driver call instead of being emulated element-by-element. The scalar and vector paths share the same underlying read/write helpers, so their semantics can't drift.
- Delete. H5Fdelete now works and cleans up both stores — the authoritative native file and the CTE cache tag — so deleting a file never leaves an orphaned tag behind.

The VFD unit suite grew to 14 sections to cover all of this: size-parity vs sec2, a concurrent SWMR reader, FAPL round-trip (cache on/off), a forced-and-verified vector path, and delete-removes-both-stores. Everything passes green end-to-end.

The CTE cache is still populate-only - writes push into it but reads always come from the authoritative native file. The FAPL toggle and this PR set up the cache plumbing; actually serving reads from CTE (a range-tracked read tier) will be done later.

### DIFF
--- a/context-transfer-engine/adapter/vfd/H5FDclio.cc
+++ b/context-transfer-engine/adapter/vfd/H5FDclio.cc
@@ -720,9 +720,12 @@ static herr_t H5FD__clio_flush(H5FD_t *_file, hid_t dxpl_id, bool closing) {
   (void)dxpl_id;
   (void)closing;
   H5FD_clio_t *file = (H5FD_clio_t *)_file;
+  // Persist the authoritative native file; fail-closed so a flush that did not
+  // reach disk never reports success. Writes are write-through, so the native
+  // file is the only store holding data to flush.
   if (file->posix_fd >= 0 && fsync(file->posix_fd) < 0) {
     H5FD_CLIO_ERROR("fsync() in flush failed");
-    return FAIL; /* never report a flush that did not persist */
+    return FAIL;
   }
   return SUCCEED;
 } /* end H5FD__clio_flush() */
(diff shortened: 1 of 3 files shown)


hyoklee (commented):
Why does it use C-style/HDF5-style comment blocks for .cc file?

mattjala:
> Why does it use C-style/HDF5-style comment blocks for .cc file?

Because that file is derived from the sec2 driver, which was in C

hyoklee:
> > Why does it use C-style/HDF5-style comment blocks for .cc file?
> 
> Because that file is derived from the sec2 driver, which was in C

This project [uses](https://github.com/iowarp/clio-core/blob/main/AGENTS.md#code-style) Google C++ Style Guide:

https://google.github.io/styleguide/cppguide.html#Comment_Style

Please use `//` only.

lukemartinlogan (approved):
I'll approve, though I agree with Joe we should try to be more conformant to overall style. We can make a new PR for deviant commenting, since these updates seem more useful overall.

Issue #768, opened 2026-07-18
Windows TCP IPC latency ~62ms/op + auto-select fastest IPC path when CLIO_IPC_MODE unset


lukemartinlogan:
## Summary

On Windows, a local `PutGet` round-trip over the **default TCP IPC path** takes **~62 ms/op** — reproducing the tick-floored latency previously seen in #620/#640. The same operation over **SHM is 331 µs (188× faster)** and over **IPC/unix-socket is 1.6 ms (38× faster)**. The default IPC transport is also the slowest by two orders of magnitude, and there is no automatic selection of a faster path when one is available.

This issue tracks two things:
1. **Investigate & fix** the TCP-path latency regression on Windows.
2. **Auto-select the fastest available IPC path** when `CLIO_IPC_MODE` is unset.

## Reproduction (native Windows, `windows-release` vcpkg build)

```
clio_run start                     # cte_config_ram.yaml
clio_cte_bench PutGet 1 1 1m 100    # 1MB blobs, x100
```

| `CLIO_IPC_MODE` | resolved | avg latency/op | aggregate BW |
|---|---|---|---|
| SHM | SHM | **331 µs** | 3018 MB/s |
| IPC | IPC | 1638 µs | 610 MB/s |
| *(unset)* / TCP | **TCP** | **62312 µs** | 16 MB/s |

## Diagnosis (in progress)

- The **62.3 ms is a fixed per-op cost** — identical at io_count 10/50/100, so it is a serialized latency, not throughput.
- **Async depth only partially hides it** (depth 16 → 33 ms), so a serialized, non-pipelineable wait sits in the round-trip.
- **62 ms ≈ 4 × 15.6 ms** (the Windows default timer tick). A `PutGet` = Put + Get = 2 round-trips, ~2 ticks each — the classic tick-floored-sleep fingerprint.
- The recv side was already hardened (`RecvZmqClientThread` uses `kZmqPollTimeoutMs = 1` via `zmq_poll`, with a code comment naming the old *"tick-floored `::Sleep` (~15.6 ms → the 40 ms TCP round-trip)"*). The residual ~62 ms implies a **second tick-floored wait remains** — prime suspects:
  - `EventManager::Wait` (`event_manager_win.cc`) `::Sleep(timeout_ms)` fallback (line ~189) when the waiter has no registered handle;
[...]

PR #769, opened 2026-07-18
feat(runtime): auto-select fastest IPC path when CLIO_IPC_MODE unset (#768)

lukemartinlogan:
## Summary

When `CLIO_IPC_MODE` is unset, `IpcManager::ClientInit` kept the member default `IpcMode::kTcp` — the **slowest** local transport was the default. This makes the client auto-select the fastest available path instead.

Measured on Windows (`windows-release` vcpkg build, `PutGet 1 1 1m 100`):

| `CLIO_IPC_MODE` | resolved | avg latency/op | aggregate BW |
|---|---|---|---|
| SHM | SHM | **331 µs** | 3018 MB/s |
| IPC | IPC | 1638 µs | 610 MB/s |
| *(unset)* → old default | **TCP** | **62312 µs** | 16 MB/s |
| *(unset)* → **this PR** | **SHM (auto)** | **~1 ms** | ~1 GB/s |

## What changed

`ClientInit` now probes for the fastest usable transport when `CLIO_IPC_MODE` is unset, in order:

1. **SHM** — `ServerInitShm` is unconditional, so a same-host runtime always has its main shared segment mapped. New `SystemInfo::SharedMemoryExists()` (a cheap open-then-close probe, no mapping) detects it. Segment present ⟺ local server up ⟺ SHM usable.
2. **IPC** — a local AF_UNIX socket the server bound for its control path (present only when the server itself runs IPC mode, or on macOS per #482). Probed via `std::filesystem::exists`.
3. **TCP** — always available; cross-host / last-resort fallback.

An explicit `CLIO_IPC_MODE` still forces that exact transport with **no probing**, so the SHM/TCP/IPC mode tests and the fallback tests (which pin SHM) are unaffected.

## Verification (live, native Windows)

- unset + same-host server → **SHM auto-selected** (62 ms → ~1 ms)
- unset + **no** server → **TCP** then a graceful "cannot connect" (no crash/hang)
- explicit `SHM`/`IPC`/`TCP` → forced verbatim, labeled `(from CLIO_IPC_MODE)`

The transport-modes "default" test previously asserted the old `kTcp` default; it now asserts the auto-selected `kShm` when a same-host server is up — the new contract.

## Scope / non-goals

This **does not fix the underlying Windows TCP latency** — that's a separate tick-floored-wait bug (62 ms ≈ 4 × the 15.6 ms Windows timer tick). The `RecvZmqClientThread` poll was already hardened (`kZmqPollTimeoutMs = 1`), but a second tick-floored wait remains in the client response path. Auto-selecting SHM sidesteps it for the common same-host case; the root-cause fix stays tracked in #768.

Closes the auto-select half of #768.

### DIFF
--- a/context-runtime/include/clio_runtime/ipc/ipc_cpu2cpu_impl.h
+++ b/context-runtime/include/clio_runtime/ipc/ipc_cpu2cpu_impl.h
@@ -10,7 +10,32 @@
 
 #include "clio_runtime/ipc/ipc_cpu2cpu.h"
 
+#include <unordered_map>
+
 namespace clio::run {
+
+#if CTP_IS_HOST
+/**
+ * Thread-local demux stash for SHM task responses (issue #768).
+ *
+ * A client thread's MPSC server (clio-<pid>-<tid>) receives the responses for
+ * EVERY task this thread submitted. When the thread has more than one async op
+ * outstanding, RecvOut(future_A) may pull future_B's response off the ring
+ * first. Blindly deserializing it into A corrupts both (a variable-size
+ * ReaddirTask read into a RenameTask slot trips GlobalDeserialize's
+ * "beyond end of data"). Park the mismatched response here, keyed by its
+ * net_key (the task's own address, stamped in SendIn), so the RecvOut that
+ * actually owns that future can claim it. Over TCP a single net-worker
+ * serializes responses, so this never surfaces there.
+ *
+ * Inline function-local thread_local => exactly one map per thread across the
+ * whole program, shared by every RecvOut<TaskT> instantiation.
+ */
+inline std::unordered_map<size_t, LoadTaskArchive> &ShmOutResponseStash() {
+  thread_local std::unordered_map<size_t, LoadTaskArchive> stash;
+  return stash;
+}
+#endif  // CTP_IS_HOST
 
 template <typename TaskT>
 Future<TaskT> IpcCpu2Cpu::SendIn(IpcManager *ipc,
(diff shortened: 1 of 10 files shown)


lukemartinlogan:
Two follow-up commits pushed alongside the auto-select feature, both from the #768 root-cause investigation:

- **`09f58c2b`** — opt-in `CLIO_WIN_TIMER_MS` knob (`timeBeginPeriod`). Off by default. Measured on the pure task round-trip bench: TCP 31.1 ms → **1.37 ms** with the daemon at 1 ms resolution (daemon-only accounts for 100% of the recovery; client-side has no effect).
- **`2855ce8a`** — `kDebug` trace points at the daemon's three TCP hand-offs (ingest / response-enqueue / drain). These produced the measured per-op trace on #768: 29 µs of work + two ~15.5 ms tick-quantized stalls, and the `wake_lane=0` receipt showing `SetNetLane` registers null lanes (called in `DivideWorkers` before `SpawnWorkerThreads` assigns lanes).

Scope note: this PR **sidesteps** the TCP latency for same-host clients via SHM auto-selection and documents/instruments the root cause. The structural fix (removing the net queue in favor of ordinary `AsyncSendIn`/`AsyncSendOut` tasks + worker-level batching) is a separate design, tracked on #768.

lukemartinlogan:
## CI finding: auto-select exposes a latent SHM-path correctness bug (blocking)

`cr_cli_cfs_rename` fails on boost(docker) and boost(ubuntu-arm); `cr_cli_cfs` times out on docker. Not flake and not resource pressure — the failing client connects fine, then its concurrent `AsyncReaddir` responses come back corrupt:

```
IPC mode = SHM (auto-selected)
GlobalDeserialize::read_binary: Attempted to read beyond end of data   (x5)
REQUIRE failed at test_cfs_rename.cc:381  Condition: n == kFiles
```

Root cause class: `ReaddirTask::entries_` is `OUT priv::vector<priv::string>` on **CTP_MALLOC (process-private heap)** — the TASK_ABI §10 category that only survives a re-serializing transport. TCP re-serializes (`SerializeOut`), so these tests always passed under the old TCP default. The FUSE/xfstests readdir path uses the *embedded* runtime (same address space). A **separate-process SHM client** running Readdir appears to have never been exercised until this PR's auto-select flipped the CFS CLI tests to SHM — and the cross-process SHM response path mangles the variable-length OUT under rename/readdir concurrency.

So the auto-select feature is doing its job (SHM is 188x faster) but it widens the set of methods that must be SHM-clean, and Readdir-class tasks currently are not.

Options:
1. **Fix the SHM marshaling for priv-heap OUT fields** (right fix; scope unknown until traced — likely the local task archive response path).
2. **Pin the CFS CLI tests to `CLIO_IPC_MODE=TCP`** (unblocks CI, but hides a real bug users would now hit by default).
3. **Constrain auto-select** (e.g. opt-in via `CLIO_IPC_MODE=AUTO`) until the SHM path is audited.

Holding for a decision before pushing anything.

Issue #774, opened 2026-07-19
Gray-Scott L=512 @ 256 ranks on Delta: writer stalls at output 1 (CTE congestion collapse: 64KB block fragmentation × ZMQ retry churn)


lukemartinlogan:
## Symptom

Gray-Scott + coeus-adapter on **Delta**, 256 processes (2 nodes × 128 ranks), clio-core 2.2:

- **L=256: works.**
- **L=512: writer stalls at output 1.** Each producer daemon ingests exactly one output's worth of data (882 MB on cn063, 1071 MB on cn064 ≈ 1 GB/node/output) and then goes flat — 0 MB growth across repeated 90 s samples, ~10 minutes parked. 91 of 128 ranks per node are sleep-polling in `hrtimer_nanosleep` (clio `Future::Wait()` backoff on CTE futures that never complete).
- **On Ares, 256 procs + L=512 works fine** — so the collapse is sensitive to the platform's network/daemon throughput balance, consistent with a congestion threshold rather than a hard functional bug.

Pipeline config (Delta): https://github.com/grc-iit/coeus-adapter/blob/iowarp_2/test/jarvis/jarvis_coeus/pipelines/delta/gray-scott-warn.yaml

## Diagnosis (code paths verified against dev)

The stall is a **live queueing collapse, not a deadlock** — mesh streams idle (heartbeat-only) while 2 of the 4 daemon worker threads spin.

**Demand side — L=512 is 8× the bytes through a scatter-everything pipeline.** Per output per node at 256 ranks: L=256 moves ~134 MB of Puts; L=512 ~1.07 GB. The engine also reads it all back doubled for derived variables (`ComputeDerivedVariables` Gets U twice and V twice). `HashBlobToContainer` (`core_runtime.cc:646-670`, unconditional blob→container hash routing; pool_query only affects container creation) sends ~50 % of both Puts and Gets to the *other* node's container with the full payload in the task.

**Service side — three size-sensitive costs:**

[...]

Issue #503, opened 2026-06-07
Re-enable temporarily-disabled CTE distributed execution validation assertions


lukemartinlogan:
## Summary

The CTE distributed integration test case **`FUNCTIONAL - Distributed Execution Validation`** (`context-transfer-engine/test/unit/test_core_functionality.cc`) has its cross-node distribution assertions **temporarily disabled** so the `distributed-test` CI workflow can be a hard (blocking) check. This issue tracks re-enabling them.

## What was disabled

The `num_nodes > 1` block that asserts work executed off node 0:

```cpp
if (num_nodes > 1) {
  REQUIRE(put_avg_completer > 0.0);
  REQUIRE(get_avg_completer > 0.0);
}
```

The Put/Get data-integrity checks in the same test still run.

## Why it fails (see #502 for the full trace)

In the 4-node docker setup every CTE blob op resolves to the client node's single local container (`completer == 0` for all 16 Puts and Gets), so the cross-node distribution assertion never holds. Root cause: the CTE pool is one-container-per-node, the test client only talks to node1, and `ResolveDirectHashQuery`'s boundary optimization (`HasContainer → Local`) keeps work on node1. This is a pre-existing test/topology mismatch, **not** a regression from the force-net / aggregation work (#501).

## Re-enable criteria

Either:
1. Make CTE blob ops genuinely fan out to remote containers (route `DirectId` to the mapped node instead of collapsing to `Local`), or
2. Rework the assertion to validate the *resolved target-node* distribution rather than the local `completer_`.

Then remove the `TEMP-DISABLED` guard in `test_core_functionality.cc` and (if needed) re-tighten the CTE step in `.github/workflows/distributed-test.yml`.

Related: #502 (root-cause analysis).

PR #775, opened 2026-07-19
fix(runtime): echo the SHM client's net_key on cross-node task responses (#774)

lukemartinlogan:
## Summary

Fixes the hard hang of **cross-node-routed tasks from `ipc_mode=shm` clients** (#774): the client's future never completed — no error, no timeout, client parked in `Future::Wait()`'s nanosleep backoff. This is the exact parked-forever state the 256-rank gray-scott writers showed on Delta (whose pipeline sets `ipc_mode: "shm"`).

## Root cause

`IpcManagerRun2Run::SendIn` repurposes `task_id_.net_key_` as its send-map key for forwarded tasks (`ipc_run2run.cc:214`), so a forwarded task's response reaches the origin daemon — and then the client — bearing the **daemon-side** key. The client's per-`net_key` response demux (bb530381 / #768) correctly refuses to deserialize a response matching no outstanding future, and parks it in the stash forever.

Timeline honesty: **before** bb530381 the blind deserialize silently tolerated the wrong key (sequential cross-node worked; concurrent was corrupt — the original #768 readdir bug). **After** it, cross-node hung outright. The missing invariant — *the response must echo the client's key* — predates both.

The ZMQ client paths already solve this exact problem: save at ingest (`ipc_cpu2cpu_zmq.cc:132`, `future_shm->client_net_key_`), restore before responding (`:325`). This PR mirrors that in the SHM path: `IpcCpu2Cpu::RecvIn` saves, `IpcCpu2Cpu::SendOut` restores. Same-node tasks never traverse run2run, so the restore is a provable no-op there.

## The diagnostic trail (details on #774)

| Control (2-node cluster, tiny 4×64KB) | Result |
|---|---|
| 1-node, SHM client | PASS |
| 2-node, TCP client, remote-routed blobs | PASS (~1 ms) |
| 2-node, SHM client, local-hash blob | PASS |
| 2-node, SHM client, remote-hash blob | **hang forever** (any size/threads/tag scope) |

gdb function tracing showed the full daemon relay chain firing (`RecvIn` → run2run response → `RecvOutCompleteOriginTask`), and client-side instrumentation caught the smoking gun: responses arriving with `got_key` = daemon task addresses vs the client's `want_key`.

## Validation (2-node deps-cpu cluster, SHM clients, ~50% cross-node scatter)

- 4×64KB, 64×64KB, 64×256KB, 64×512KB per-output phases: **all complete** (~1.5 GB/s put throughput at 0.5 MB blobs). Previously the first cross-node-routed op hung forever.
- Mismatch tracing post-fix shows only genuine out-of-order sibling responses transiting the stash (claimed by their own waiters) — the demux working as designed.
- 64×4MB phase: one node completed all outputs at 2.4 GB/s; the other stalled in output 1 — that residual is the **size-dependent congestion layer** #774 originally described (L=256 vs L=512), now investigable on top of working cross-node ops. Tracked separately on the issue.

## CI gate (why this survived so long, and why it can't again)

No suite combined **SHM client × multi-node × client-origin cross-node routing**: cross-node blob assertions were disabled (#503), and every distributed suite uses TCP clients or the CFS adapter. This PR adds a hard gate — `grayscott_congestion/run_tests.sh tiny` (2-node docker, SHM client, cross-node-routed Puts/Gets must complete) — wired into `cluster-tests.yml`, plus the full gray-scott congestion repro harness (`GS_*` knobs, per-blob tracing, watchdog, size sweep).

Issue #774.

### DIFF
--- a/.github/workflows/cluster-tests.yml
+++ b/.github/workflows/cluster-tests.yml
@@ -216,6 +216,20 @@ jobs:
           export HOST_UID="$(id -u)" HOST_GID="$(id -g)"
           ./context-transfer-engine/test/integration/cfs_distributed/run_tests.sh
 
+      # Cross-node SHM-client gate (#774): direct CTE-core Puts/Gets from an
+      # ipc_mode=shm client whose blob names HashBlobToContainer-route to the
+      # OTHER node's container. This exact combination (SHM client x multi-node
+      # x client-origin cross-node routing) had ZERO coverage: it hung forever
+      # (run2run repurposes task_id_.net_key_, the response returned bearing
+      # the daemon's key, and the client demux parked it) and no suite noticed
+      # because cross-node blob assertions were disabled (#503) and every other
+      # distributed test used TCP clients or the CFS adapter. Hard gate.
+      - name: CTE cross-node SHM client gate (2-node, issue #774)
+        if: ${{ !cancelled() }}
+        run: |
+          export HOST_UID="$(id -u)" HOST_GID="$(id -g)"
+          ./context-transfer-engine/test/integration/grayscott_congestion/run_tests.sh tiny
+
       - name: CTE FUSE test (mount + POSIX I/O)
         if: ${{ !cancelled() }}
         run: |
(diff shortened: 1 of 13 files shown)


lukemartinlogan:
Now carries the complete #774 fix family (5 fixes — see the issue's final comment for the full diagnostic trail). Integrated validation: **12/12 consecutive full size sweeps clean** vs a ~60–75% pre-fix stall rate per sweep; the deterministic 45s-partition test returns bounded errors instead of hangs.

PR #776, opened 2026-07-19
637 gpu vector compressed transfer

rajnipawar:


lukemartinlogan (changes requested):
* Remove all pdf and png files
* Remove all custom py files
* Remove the "results" directory
* Remove patch files, you should be able to just augment the code with macros if anything. 
* Fix merge conflicts

lukemartinlogan:
* Remove markdown files describing preliminary results. Keep markdowns that are README
* Add a docker cluster unit test, right now I'm only seeing slurm which is probably delta-specific. A docker cluster is more portable and I can verify on my local machine to some extent.

lukemartinlogan:
We will be better integrating this code with the current dev branch in a separate issue. A new PR will be made hopefully tomorrow.

PR #778, opened 2026-07-20
feat(GPUH5): Make kernel persistent, fix async double buffering, and optimize CLIO CTE

abscosmos:
Will list out changes to CLIO in a bit. Still need to add more unit tests for those changes. This PR also allows importing/exporting HDF5 files from GPUH5.


abscosmos:
@lukemartinlogan The first failing test is flaky, and the second one is a preexisting issue. Can you merge this so in the AD I can link to this repo instead of my fork?

lukemartinlogan (approved):

Issue #787, opened 2026-07-22
Measure SHM metadata cache reader retry/fallback rate under sustained write load (#783 follow-up)


lukemartinlogan:
## Summary

The shared-memory metadata cache from #783 lets clients read CTE metadata and node-local RAM-resident payloads with zero IPC. Reads are optimistic (seqlock): a reader retries if a writer was mid-update, and **falls back to the RPC path** once it exhausts its retry budget.

**We have never measured the reader retry rate under realistic write load.** If it is high for hot blobs, readers silently drop to RPC and the speedup quietly disappears — with no error and no log.

## Why this is not already answered

The Phase 2 torture test (`test_lease_exec`, `[torture]`) reported ~90% retries:

```
writes=503,400   read_ok=25,554,659   retries=242,412,642   torn=0   steals=1
```

That number is **not evidence of a problem**: that writer deliberately holds its update window open almost continuously to widen the tear window, which is the opposite of realistic. But it is equally **not evidence of safety** — it tells us nothing about the real ratio.

## What to measure

For a blob under sustained rewrite, while concurrent clients read it:

1. Reader **retry rate** (retries per successful read).
2. **Fallback rate** — how often `TryGetBlobRecordShm` / `TryReadBlobShm` return false and the caller takes RPC.
3. Effective end-to-end latency at the API level (`Tag::GetBlobSize`, `Tag::GetBlob`) under that load, versus the quiescent numbers already recorded (metadata ~0.15 µs, payload ~0.21 µs, vs ~84–143 µs for RPC).

Sweep write rate and reader count; the interesting region is where write frequency approaches read frequency on the same key.

## Why it matters

The fast path is designed to degrade to RPC rather than ever return wrong data, which is the correct trade. The risk is purely that the degradation is **invisible**: a workload could be getting no benefit at all and nothing would report it.

## Possible follow-ups depending on results

- Expose fallback/retry counters so the degradation is observable rather than silent.
- If a hot key can starve readers, consider a bounded lease on the read side (`Lease` already exists, `clio_ctp/memory/smart_ptr/lease.h`) so a reader can make progress instead of spinning and giving up.
- Consider double-buffering a record so a writer never invalidates in place.

## Context

- Design doc: `project_cte_shm_metadata_cache.md`, §5.3b and Phase 2 notes.
- Flagged since Phase 2 and carried through to completion of #783 without being closed out.

Issue #786, opened 2026-07-22
clio_run_thrpt_bench --test-case bdev_allocation segfaults against an external daemon


lukemartinlogan:
## Summary

`clio_run_thrpt_bench --test-case bdev_allocation` segfaults in the **client** shortly after `CLIO_INIT`, when run against an **externally started** runtime (`CLIO_WITH_RUNTIME=0`).

It reproduces on **both** the RAM and file bdev paths, so it is not specific to a bdev transport.

## Reproduction

```bash
# terminal 1
./clio_run start

# terminal 2
export CLIO_WITH_RUNTIME=0 CLIO_IPC_MODE=SHM

# RAM bdev
./clio_run_thrpt_bench --test-case bdev_allocation --threads 1 --duration 3 \
    --output-dir ram --max-file-size 256m
# -> Segmentation fault (core dumped), exit 139

# file bdev — same crash
./clio_run_thrpt_bench --test-case bdev_allocation --threads 1 --duration 3 \
    --output-dir /tmp/clio_bench --max-file-size 256m
# -> Segmentation fault (core dumped), exit 139
```

## Observed behaviour

Client output stops immediately after the bdev-type banner:

```
Lane policy: map_by_pid_tid (default)
Using RAM-based BDev            <- (or "Using file-based BDev: ...")
Segmentation fault (core dumped)
```

The last thing printed is the banner emitted just *before* `bdev_client.AsyncCreate(...)` in `clio_run_thrpt_bench.cc`, so the crash is in or immediately after the bdev `Create` round-trip.

The **daemon survives** and stays healthy — it keeps servicing other clients and logging normally, so this is a client-side fault, not a runtime crash.

## Scope

- Affects `--test-case bdev_allocation`. Other cases in the same binary (`latency`) run fine against the same external daemon and produce results.
- Independent of `bdev_type`: `ram` and `file` both crash identically.
- Only observed with an external daemon (`CLIO_WITH_RUNTIME=0`).

## Not caused by #783

Found while testing the shared-memory metadata cache (#783). Ruled out as a regression from that work:

- The **file** bdev path crashes identically, and #783 touched only the `kRam` transport path.
- The crash occurs before any #783 code is reachable in this test case.

Recorded there so it would not be silently absorbed into that issue.

## Suggested next steps

[...]

PR #788, opened 2026-07-22
CTE shared-memory metadata cache: zero-IPC client reads (#783)

### LINKED ISSUE #783: CTE: shared-memory metadata cache + SHM RAM bdev for zero-IPC synchronous blob reads
## Motivation

Blobs are arbitrary-sized, so we cannot cache them like fixed-size file pages. Every small-I/O metadata lookup currently costs a runtime round-trip. Measured on a 6-core WSL host (issue #781 branch, `clio_run_thrpt_bench --test-case latency`):

| Transport | 1 thread | 8 threads |
|---|---|---|
| SHM | ~71 µs | ~240 µs |
| IPC (unix socket) | ~125 µs | ~297 µs |
| TCP (loopback) | ~472 µs | ~692 µs |

Even the best case is ~71 µs per op, and it degrades under concurrency. There is no hardware extension that rescues this. The only way to make small reads cheap is to **not make the round trip at all**.

## Goal

Put the CTE's metadata maps and RAM-bdev data in shared memory so client threads can serve reads **synchronously, in-process, with zero IPC**.

- `string -> TagId`, `TagId -> TagInfo`, `string -> BlobId`, `BlobId -> BlobInfo` live in shared memory.
- The **runtime is the sole owner/writer**. Clients are **read-only**.
- Reads become synchronous SHM lookups. **Writes stay fully asynchronous (write-through).**
- Any bdev that is both node-local and `kRam` becomes SHM-backed, so clients can read blob payloads directly.

## Planned work

1. **Runtime-wide metadata segment.** New 8 GB fixed-size, non-pre-faulted SHM segment (`kMetadataSegment`), created at context-runtime startup, owned by the runtime as a whole (not CTE-specific). `IpcManager` collects and exposes it alongside the existing `kMainSegment`.
2. **SHM container support.** We have `ctp::ipc::vector` and `ShmContainer<AllocT>`, but **no `ipc::string` and no `ipc::unordered_map`**. Both need to be written.
3. **SHM analogs of `TagInfo` / `BlobInfo`.** These are non-trivial: they currently hold `priv::string`, `priv::vector<BlobBlock>`, and a `ctp::Mutex`, and are stored in the maps as `std::shared_ptr`.
4. **Create-time propagation.** CTE `CreateTask` carries the map locations; the CTE client stores them and attaches the segment read-only.
5. **Client `GetBlob` fast path.** Resolve via the SHM maps; if every block is node-local `kRam`, copy directly from the SHM bdev. Otherwise fall back to the existing async RPC path.
6. **SHM-backed RAM bdev**, mappable read-only into client processes.

## Design problems that must be solved first

These are the parts that make this a substantial refactor rather than a mechanical port. Each one can silently corrupt data if handled naively.

**1. Reclamation without `shared_ptr`.** The maps currently hold `std::shared_ptr<TagInfo>` / `std::shared_ptr<BlobInfo>` (61 call sites), and `core_runtime.h` explicitly relies on this: *"a concurrent erase just drops the map's reference while any in-flight handle keeps the object alive -- no use-after-free."* Shared memory has no `shared_ptr`. Readers are in **other processes** and can be `SIGKILL`ed mid-read. We need epoch-based reclamation with per-client epoch slots published in SHM, plus liveness detection so a dead client cannot wedge reclamation forever.

**2. Readers must never block the runtime.** Any scheme where the runtime waits on a client-held lock is unacceptable — a killed client would wedge the runtime permanently. All client reads must be optimistic (seqlock: read sequence, read payload, re-validate sequence, retry). **Consequence: clients cannot reuse `unordered_map_ll`** — its `find()` takes locks, i.e. it *writes*. A separate lock-free read-only lookup over the same memory layout is required.

**3. Coherence against the DataOrganizer.** The frecency/tiering organizer actively moves and evicts blobs. A client can read block offsets, then have the runtime free or reuse those blocks before the copy completes — yielding silently wrong data with no error. Needs a per-blob generation validated before *and* after the payload copy, and epoch-deferred reuse of RAM-bdev blocks.

**4. Read-your-own-writes.** "Writes fully async" breaks RYOW: a client that does `PutBlob` then `GetBlob` may not observe its own write. Needs a per-client pending-write set that forces the slow path for affected keys until the write is acknowledged.

**5. `PROT_READ` enforcement.** Every client-side path must be genuinely non-mutating. A stray write to a read-only mapping is a `SIGSEGV`, not a wrong answer. This rules out lock acquisition, lazy initialization, and refcounting anywhere on the read path.

**6. Trust boundary (needs a decision).** A shared metadata segment means *any* client can read *every* tag and blob name/size on the node. Today a client only sees what the runtime chooses to return. This is a real change in the isolation model and should be an explicit, recorded decision rather than a side effect.

**7. `/dev/shm` sizing.** An 8 GB sparse segment maps fine, but on a default container/WSL `/dev/shm` (64 MB here) touching beyond that limit is a `SIGBUS`, not an allocation failure. We need a graceful "metadata cache full -> fall back to RPC" path and documented tmpfs sizing for realistic testing.

## Phasing

Dual-write is the key de-risking step: the existing structures stay authoritative until the SHM mirror is proven, so a 127-call-site refactor never has a flag-day.

- **Phase 0** — Benchmark harness for CTE metadata ops; record the baseline this must beat.
- **Phase 1** — `kMetadataSegment` + `IpcManager` plumbing.
- **Phase 2** — `ipc::string`, `ipc::unordered_map` + standalone unit tests (incl. concurrent reader/writer torture).
- **Phase 3** — `ShmTagInfo`/`ShmBlobInfo`; runtime **dual-writes** both old and new structures. Old path stays authoritative.
- **Phase 4** — Propagate map locations via `CreateTask`; client attaches read-only.
- **Phase 5** — Client fast path for **metadata-only** ops first (`GetBlobSize`, `GetBlobScore`) — lower blast radius than payload reads.
- **Phase 6** — SHM-backed RAM bdev + client payload fast path.
- **Phase 7** — Retire the dual-write and the legacy structures.

## Acceptance

- Client `GetBlob` for a node-local RAM-resident blob completes with **zero IPC round-trips**, target **< 5 µs** (vs ~71 µs today).
- Concurrent reader/writer torture test passes with clients killed at random mid-read.
- Correctness parity: existing CTE test suites pass unchanged at every phase.

lukemartinlogan:
Closes #783.

Client reads of node-local CTE metadata and blob payloads now complete with **zero IPC**, by reading shared-memory structures the runtime owns instead of making a round-trip.

Draft: the implementation is complete and tested, but see **Open questions** below before merging.

## Why

Blobs are arbitrary-sized, so the fixed-size page-cache trick does not apply, and every metadata lookup was a runtime round-trip. Measured on a 6-core WSL host before any of this work:

| Transport | 1 thread | 4 threads | 8 threads |
|---|---|---|---|
| SHM | 70.6 µs | 145.3 µs | 240.5 µs |
| IPC (unix socket) | 124.5 µs | 175.5 µs | 297.1 µs |
| TCP (loopback) | 471.7 µs | 574.8 µs | 691.5 µs |

The floor was ~71 µs and it degraded under concurrency. No hardware extension fixes that; the only way to make small reads cheap is to not make the trip.

## Results

**Read-only (steady state, cached blobs — read-mostly workloads):**

| Read type | SHM | RPC | Speedup |
|---|---|---|---|
| Metadata | 0.14–0.21 µs | 89–102 µs | ~430–730x |
| Payload (4 KB) | 0.20–0.31 µs | 110–140 µs | ~350–700x |

**Write-then-read cycle (write-dominated — quote this one for read-after-write):**

| Measurement | SHM | RPC | Ratio |
|---|---|---|---|
| write alone | 79–140 µs | (same) | — |
| write + metadata read | 81–114 µs | 162–195 µs | **~2x** |
| write + payload read | 83–128 µs | 187–278 µs | **~2x** |

Writes are deliberately **not** accelerated here, so a read-after-write caller sees ~2x, not several-hundred-x. The read component becomes effectively free (~0.2 µs); Amdahl caps the rest. Both figures are recorded so the right one gets quoted for the right workload.

The RPC baseline swings run-to-run on this host (89–280 µs) while the SHM path stays ~0.2 µs, so these are ranges, not point estimates.

## Design

Full write-up in `project_cte_shm_metadata_cache.md` (in this PR). The load-bearing constraint is that **a client may be `SIGKILL`ed at any instant, including mid-read**, so nothing the runtime does may block on a client. That rules out the obvious designs and drives everything below.

- **Runtime-wide metadata segment** (`kMetadataSegment`), 8 GB sparse, created at startup, clamped to half of RAM. Non-fatal throughout: no segment means clients simply use RPC.
- **`ipc::string`** — offset-based, 23-byte SSO. Uses FNV-1a rather than `std::hash`, which is stable neither across processes nor libstdc++ versions; a shared table keyed on it would silently corrupt.
- **`ipc::unordered_map`** — open addressing, per-slot seqlock, tombstones. **Never rehashes**: growth is the one operation that cannot be made safe for untracked cross-process readers, since a rehash frees a table out from under them. Capacity is fixed and a full table degrades to RPC.
- **`TimedMutex`** — `ctp::Mutex` is a ticket lock and cannot be reclaimed (a dead holder never advances `head_`). This is an unfair CAS lock that trades FIFO fairness for reclaimability, using PID **plus process start time** since PIDs recycle.
- **`Lease` / `SeqRead`** — `gen_` is correctness, `lease_mutex_` is progress/lifetime. A seqlock reader stores nothing, so a reader that dies leaves nothing to reclaim; a lease is the only client-side construct whose abandonment needs servicing, so the design minimises how often one is held.
- **SHM-backed RAM bdev** — mapped as one contiguous sparse region rather than a page table; segment name derives from (runtime pid, pool id) so clients reconstruct it without anything being published.
- **Cache, never authority.** Clients map read-write (acquiring a lease is a store), so a corrupted cache must stay droppable. The runtime keeps its own structures permanently.

## Correctness properties under test

- **Read-your-own-writes** — a waited `PutBlob` is visible to the next read.
- **Cached answers agree with RPC** — a fast wrong answer is worse than no fast path.
- **A miss is "not cached", never "absent"** — always falls back, never reports missing data.
- **Torn reads rejected** — seqlock validated before and after.
- **`placement_gen_` validated before AND after the payload copy** — the metadata seqlock protects the record, but the real hazard is the DataOrganizer relocating blocks mid-`memcpy`.
- **Kill-readers torture test** — 6 seqlock readers + 2 lease holders SIGKILLed at random against a writer: 503k writes, 25.5M cross-process reads, **0 torn reads**, abandoned lease reclaimed.

## Testing

All on a from-scratch rebuild (`build-cpu` destroyed, 163 targets, 0 errors):

| Suite | Result |
|---|---|
| `test_string_exec` | 43 assertions / 10 cases |
| `test_unordered_map_exec` | 6191 assertions / 11 cases |
| `test_lease_exec` (incl. torture) | 28 assertions / 5 cases |
| `test_cte_shm_metadata_cache` | 7/7 |
| `test_core_functionality` | 13/13 |
| `clio_run_bdev_chimod_tests` | 21/21 |
| Transports (latency) | SHM 81.9 µs · IPC 150.0 µs · TCP 573.7 µs |

## Open questions for review

1. **Reader retry/fallback rate under concurrent write load is unmeasured** (#787). The fast path degrades to RPC rather than ever returning wrong data — correct, but the degradation is currently **invisible**. A workload could get no benefit and nothing would say so. Exposing fallback counters is the obvious follow-up.
2. **Fixed capacities** (64K tags, 256K blobs per pool) are chosen, not derived. Exceeding them degrades to RPC silently, same visibility gap.
3. **Node-wide metadata visibility** was accepted as a single-tenant assumption: any client can read every tag/blob name and size on the node. Fine today, but it is a real change to the isolation model and should be a conscious call.
4. **`MetadataDirectory` caps at 32 registered pools.** Beyond that a pool simply is not discoverable (degrades to RPC).

## Notes

- Includes a merge of `origin/dev` to pick up PR #782, which the design depends on.
- #786 was filed during this work and **closed as invalid** — it was ABI skew from my own partial rebuilds, not a repo defect. Worth knowing: changing the layout of a widely-included type and rebuilding only some targets produces crashes that look like unrelated bugs elsewhere.

### DIFF
--- a/context-runtime/include/clio_runtime/config_manager.h
+++ b/context-runtime/include/clio_runtime/config_manager.h
@@ -382,7 +382,12 @@ class ConfigManager : public ctp::BaseConfig {
   // costs only the pages actually touched. NOTE: on a host whose /dev/shm is
   // smaller than the live set, touching past the tmpfs limit raises SIGBUS,
   // not ENOMEM. See CalculateMetadataSegmentSize().
-  size_t metadata_segment_size_ = ctp::Unit<size_t>::Gigabytes(8);
+  // 0 means "auto-calculate" — CalculateMetadataSegmentSize() supplies the
+  // default. Initialising this to a non-zero size would make the explicit
+  // override branch there permanently true and the documented sentinel a lie;
+  // it only worked because LoadDefault() resets it to 0. Set via the
+  // `metadata_segment_size` key under `runtime` in the config file.
+  size_t metadata_segment_size_ = 0;
 
   u32 port_ = 9413;
   // If true (CLIO_EPHEMERAL / --ephemeral), skip the default compose at startup.
--- a/context-runtime/include/clio_runtime/config_manager.h
+++ b/context-runtime/include/clio_runtime/config_manager.h
@@ -175,6 +175,17 @@ class ConfigManager : public ctp::BaseConfig {
   size_t CalculateQueueSegmentSize() const;
 
   /**
+   * Calculate the runtime-wide metadata segment size (issue #783).
+   *
+   * Backs the CTE shared-memory metadata cache. The reservation is large and
+   * sparse: it is never pre-faulted, so untouched pages cost nothing. Sizing is
+   * therefore about the address-space reservation, not RAM.
+   *
+   * @return Calculated size in bytes, or the explicit size if one is configured
+   */
+  size_t CalculateMetadataSegmentSize() const;
+
+  /**
    * Get memory segment size
    * @param segment Memory segment identifier
    * @return Size in bytes
@@ -366,6 +377,12 @@ class ConfigManager : public ctp::BaseConfig {
 
   size_t main_segment_size_ = ctp::Unit<size_t>::Gigabytes(1);
   size_t client_data_segment_size_ = ctp::Unit<size_t>::Megabytes(256);
+  // issue #783: metadata segment. Deliberately huge and never pre-faulted --
+  // shm_open + ftruncate + mmap is lazily populated, so the 8 GB reservation
+  // costs only the pages actually touched. NOTE: on a host whose /dev/shm is
+  // smaller than the live set, touching past the tmpfs limit raises SIGBUS,
+  // not ENOMEM. See CalculateMetadataSegmentSize().
+  size_t metadata_segment_size_ = ctp::Unit<size_t>::Gigabytes(8);
 
   u32 port_ = 9413;
   // If true (CLIO_EPHEMERAL / --ephemeral), skip the default compose at startup.
@@ -377,6 +394,7 @@ class ConfigManager : public ctp::BaseConfig {
   std::string main_segment_name_ = "chi_main_segment_${USER}";
   std::string client_data_segment_name_ = "chi_client_data_segment_${USER}";
   std::string queue_segment_name_ = "chi_queue_segment_${USER}";
+  std::string metadata_segment_name_ = "chi_metadata_segment_${USER}";
 
   // Networking configuration
   std::string hostfile_path_ = "";
--- a/context-runtime/include/clio_runtime/ipc_manager.h
+++ b/context-runtime/include/clio_runtime/ipc_manager.h
@@ -337,6 +337,17 @@ class IpcManager {
   /** Client-side: record the directory offset learned from ClientConnect. */
   void SetMetadataDirOffset(u64 off) { metadata_dir_off_ = off; }
 
+  /**
+   * PID of the runtime this process talks to (its own pid when this IS the
+   * runtime). Clients use it to reconstruct per-pool SHM segment names, e.g.
+   * a RAM bdev's data segment (issue #783), without those names having to be
+   * published anywhere.
+   */
+  int GetRuntimePid() const {
+    return runtime_pid_ != 0 ? runtime_pid_
+                             : static_cast<int>(ctp::SystemInfo::GetPid());
+  }
+
   /** Pid-based allocator ids of this runtime's SHM segments (pid.1 main,
    *  pid.2 queue, pid.3 metadata). Reported to clients via ClientConnect so
    *  they attach the right allocators instead of assuming (1,0)/(2,0). */
--- a/context-runtime/include/clio_runtime/ipc_manager.h
+++ b/context-runtime/include/clio_runtime/ipc_manager.h
@@ -302,12 +302,26 @@ class IpcManager {
     return main_allocator_;
   }
 
+  /**
+   * Returns the runtime-wide metadata allocator (issue #783), or nullptr if
+   * the metadata segment is not attached. Callers MUST null-check: the segment
+   * is optional, and a client that failed to attach it simply falls back to
+   * the RPC path rather than failing.
+   */
+  CLIO_TASK_ALLOC_T *GetMetadataAllocator() { return metadata_allocator_; }
+
+  /** True if the metadata segment is available for shared-memory caching. */
+  bool HasMetadataSegment() const { return metadata_allocator_ != nullptr; }
+
   /** Pid-based allocator ids of this runtime's SHM segments (pid.1 main,
-   *  pid.2 queue). Reported to clients via ClientConnect so they attach the
-   *  right allocators instead of assuming (1,0)/(2,0). */
+   *  pid.2 queue, pid.3 metadata). Reported to clients via ClientConnect so
+   *  they attach the right allocators instead of assuming (1,0)/(2,0). */
   ctp::ipc::AllocatorId GetMainAllocatorId() const { return main_allocator_id_; }
   ctp::ipc::AllocatorId GetQueueAllocatorId() const {
     return queue_allocator_id_;
+  }
+  ctp::ipc::AllocatorId GetMetadataAllocatorId() const {
+    return metadata_allocator_id_;
   }
 
   /**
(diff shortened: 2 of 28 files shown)


lukemartinlogan:
## ⚠️ The metadata segment does not map on Windows — the feature is silently disabled there

Found while auditing CI flakes. On every Windows job, the 8 GB metadata segment fails to create and the runtime falls back:

```
ipc_manager.cc:1004 WARNING ServerInitShm Metadata segment: requested 8589934592 bytes
                            exceeds half the memory ...
ipc_manager.cc:1012 INFO    ServerInitShm Initializing metadata shared memory segment:
                            8587180032 bytes (8189 MB)
MapViewOfFile failed with error: Access is denied.
posix_shm_mmap.h:113 ERROR  shm_init Failed to ...
ipc_manager.cc:1023 WARNING ServerInitShm: metadata segment
                            'chi_metadata_segment__9413' unavailable
```

### Why "Access is denied" and not an out-of-memory error

`SystemInfo::MapSharedMemory` (`system_info.cc:675`) calls:

```cpp
MapViewOfFile(fd.windows_fd_, FILE_MAP_ALL_ACCESS, highDword, lowDword, size);
```

`MapViewOfFile` returns `ERROR_ACCESS_DENIED` when the requested view is **larger than the underlying file-mapping object**. So `CreateFileMapping` did not back the full ~8 GB — plausible on a 16 GB runner, where a pagefile-backed 8 GB section hits the commit limit — and the subsequent full-size view request is then rejected with a misleading error.

The RAM clamp is not saving us here: half of 16 GB is 8 GB, so it clamps 8589934592 → 8587180032, still ~8 GB.

### Why this matters more than a warning

The fallback is deliberate and non-fatal — that was the design. The consequence is that **on Windows this PR's entire feature is inert**, while CI stays green. Every Windows job is exercising the pre-existing RPC path, not the SHM metadata cache. So the Windows column of this PR proves only that the fallback works, not that the feature does.

That is the same vacuous-green trap hit earlier in this branch's history with `cte_core_functionality_simple_tests`, which passed 9/9 while never creating a CTE pool.

### Suggested follow-ups before this leaves draft

1. **Reserve rather than commit on Windows.** Create the section with `SEC_RESERVE` and commit pages on demand, which matches the "should not be pre-faulted" requirement in the design and avoids the commit-limit failure entirely.
2. **Clamp the request to something a 16 GB runner can back**, or make the capacity configurable so CI can request a small segment.
3. **Make the coverage non-vacuous**: have at least one test assert `HasMetadataSegment()` is true and fail loudly if the cache is unavailable, so a silent fallback cannot read as a pass. Gate it so platforms that legitimately cannot provide the segment skip rather than silently no-op.
4. Fix the error reporting so a too-large view is not reported as "Access is denied".

I have not fixed any of this yet — recording it so the Windows results on this PR are not misread as validating the feature.

Issue #781, opened 2026-07-21
Anti-deadlock scheduler: guarantee progress under mislabeled/blocking tasks


lukemartinlogan:
## Motivation — anti-deadlock scheduling

The runtime's liveness currently depends on tasks being *well-behaved*. A single **non-yielding** task (a CPU spin, a blocking syscall, or an infinite loop inside a coroutine that never returns control) permanently wedges the worker it lands on. Because workers are coroutine-based, an *expensive-but-cooperative* task (one that yields) is already fine — the lethal case is the **non-cooperative** task. Today:

- `DefaultScheduler::ClientMapTask` returns `0` for everything (`num_sched_queues==1`), so **every** externally-ingressed task funnels onto **lane 0 / the scheduler worker**. A few mislabeled blocking tasks there starve the whole ingress path.
- The worker pool is **fixed** at `num_threads` (default 4); there is **no** dynamic growth, **no** work-stealing, and `Scheduler::RebalanceWorker` is a declared-but-never-called no-op.
- The only routing signal is `TaskStat.io_size_ >= 4096` (round-robin to I/O workers) plus a learned per-method CPU model. **There is no `TASK_BLOCKING`/`TASK_LONG_RUNNING` label** — cheap-vs-expensive is inferred, and a task that is cheap by `io_size_`/`compute_` but actually blocks routes straight to lane 0 and starves it.

**This is the likely cause of the Delta hangs.** We should not stake the safety of the runtime on tasks being honestly labeled.

**Goal:** an algorithm that runs a group of *arbitrarily complex* operations with the **minimum number of threads**, while keeping **low p99 latency for quick tasks** and **scaling for large tasks** — and that *guarantees forward progress even for adversarial/mislabeled tasks*.

## Design principles

1. **Never trust labels for safety.** Correctness (liveness) comes from a monitor that detects a stalled worker and routes around it — not from tasks declaring their cost.
2. **Route on *measured* load, not predictions.** Mapping uses real-time measured worker load; we explicitly do **not** use a learned cost PDF to route (predictions are unstable and a mislabel poisons them). The perf-bin histogram is kept for **telemetry only**.
3. **Elastic, work-conserving, minimal threads.** Start at the baseline pool; grow *only* on demonstrated stall/starvation; retire extra threads after an idle cooldown (hysteresis).
4. **Isolation over prevention.** A pathological task strands *one* thread, never the runtime. Its backlog is stolen and continues.

## Two layers

[...]

Issue #705, opened 2026-07-08
Awaiter resumes before bdev handler finishes its output stores when the handler yields (POSIX AIO poll loop) - short reads/writes on file bdev in docker CI


hyoklee:
## Symptom

In the `boost (docker deps-cpu)` CI job, blob I/O against the **file bdev** intermittently reports short transfers:

```
core_runtime.cc ERROR ReadData: READ FAILED - task[0] read 0 bytes, expected 65536
```

This fails `cte_tiered_dram_default_all` deterministically-ish (dev runs 28884611044, 28837814275, and PR #704 runs) and makes `cte_tag_force_net` ("Tag - Large Data PutGet") flake 2/3 before `--repeat until-pass:3` rescues it. RAM-tier I/O never fails.

## Evidence that this is a premature-resume race, not an I/O failure

From run 28910505466 (PR #704, which added rc logging):

```
ReadData: READ FAILED - task[0] read 0 bytes, expected 65536
ReadData: READ FAILED - task[0] read 65536 bytes, expected 65536   <-- !!
```

The second line is logically impossible unless `task->bytes_read_` **changed between the failure check and the error log line** (`core_runtime.cc` `ReadData`, the `if (task->bytes_read_ != expected_size)` block): the check saw a stale value, the print saw the final one. The awaiter of `AsyncRead` was resumed **before the bdev handler finished writing its output fields**.

## Why only the file bdev, and only in docker

- The container's seccomp profile blocks `io_uring_queue_init`, so `FsBdevTransport` falls back to **POSIX AIO** (see `OpenBackingFile` in `fs_bdev_transport.cc`).
- POSIX AIO completions are slow enough that `ReadBlocks`/`WriteBlocks` actually enter their poll loop: `while (!IsComplete(token, result)) CLIO_CO_AWAIT(clio::run::yield(10.0));`
- The **mem bdev transport completes synchronously** (no yield) and never exhibits the failure; native runners complete io_uring ops fast enough that the first `IsComplete` usually succeeds.
- So the trigger is: **a bdev task that suspends (yields) mid-handler**. When it does, its awaiter can be resumed before the handler's final stores (`bytes_read_` / `bytes_written_`, and the staging memcpy for device buffers) are visible/executed.

[...]

Issue #680, opened 2026-07-03
clio FUSE: xfstests generic failures — feature gaps, timestamp bugs, and daemon hangs


lukemartinlogan:
## Summary

Catalog of `generic/*` xfstests that **fail** against the clio FUSE filesystem, to drive feature work. Established by running the entire `generic` group (793 tests) against `clio_cte_fuse`:

| Outcome | Count |
|---|---|
| pass (max passing set, now the CI gate) | 50 |
| notrun (require `SCRATCH_DEV`, which the fs can't provide) | 702 |
| **FAIL** | 35 |
| **HANG** (daemon wedges into D-state) | 5 |

Of the 91 tests that actually execute, 50 pass. The 50 passing tests are gated in CI (`scripts/xfstests/ci_baseline_pass.txt`, issue #677); the 40 below are **not** run in CI and are tracked here. Method: discovery run + 2 verification runs; `generic/013` is flaky (fsstress-driven) and excluded from the gate.

### 1. Hard hangs — FUSE daemon wedges (5) 🔴 highest priority
Daemon stops servicing I/O; client processes get stuck in uninterruptible **D-state** and only a lazy/force unmount recovers them. Each wedges an unattended run entirely.
`generic/089` `generic/100` `generic/208` `generic/323` `generic/438`
- `generic/438` reproduces cleanly: `t_mmap_fallocate` (mmap + fallocate) wedges the daemon.

### 2. Direct I/O (O_DIRECT) — EIO / broken hole reads (11)
O_DIRECT writes return `EIO`; direct reads from holes return wrong data. `generic/091` additionally reports the fs supports **no** `fallocate` mode.
`generic/014` `generic/074` `generic/075` `generic/091` `generic/112` `generic/133` `generic/198` `generic/247` `generic/465` `generic/647` `generic/729`

### 3. Timestamps not updated — mtime/ctime/atime (8)
Writes, mapped writes, and truncates don't advance mtime/ctime; atime handling and pre-epoch timestamps also fail.
`generic/035` `generic/080` `generic/192` `generic/215` `generic/221` `generic/258` `generic/309` `generic/313`

### 4. FUSE daemon crash under stress — "Transport endpoint is not connected" (4)
Daemon dies mid-test under fsx op streams / copy_range.
`generic/127` `generic/363` `generic/521` `generic/522`

### 5. POSIX file locking (1)
fcntl locks not surfaced (`lock info not found` in `/proc/locks`).
`generic/504`

[...]

Issue #642, opened 2026-06-27
Concurrent data structures + Lightbeam SHM MPSC transport redesign


lukemartinlogan:
## Goal
Reduce locking granularity by protecting key data structures against concurrent
modification, and redesign the Lightbeam SHM transport into a **named,
multi-producer / single-consumer (MPSC)** buffer so the runtime no longer
allocates per-task FutureShm in shared memory.

### 1. `mcsp_unordered_map`
- New concurrent map structured like `unordered_map_ll`, protected against
  concurrent modification (fine-grained / per-bucket locking).

### 2. Lightbeam SHM transport rewrite (named transfer buffer, MPSC)
- Created with a unique SHM name; server allocates one segment via the SHM
  `MemoryBackend` (default 128KB transfer space).
- Header layout:
  - `ShmXferHeader { u64 conn_id_; u32 xfer_off_, xfer_size_, rem_off_, rem_size_; hipc::atomic<bool> ready_; }`
  - `ShmTransportHeader { hipc::atomic<u64> head, tail, connection_id, xfer_id_head, xfer_id_tail; int pid, tid; size_t max_capacity; ShmXferHeader xfers_[256]; }`
- Client connects to the named SHM; bumps `connection_id` for its id.
- `Send / RecvMetadata / Recv` structured as today; add `DONTWAIT` flag (Recv
  returns immediately unless a transfer is in progress).
- `SendBytes(data, size)`: chunk into <=32KB pieces over the ring (xfer_id_tail
  fetch_add + bounded 256 in-flight; tail fetch_add for ring offset; wait for
  ring capacity with server-liveness timeout; memcpy; fill `xfers_[xfer_id]`;
  mark ready).
- `RecvBytes(data, size)`: reverse — wait for xfer_id, wait for `ready` with
  connection-liveness timeout (skip+mark-dead after ~1s; track connections in an
  unordered_map), memcpy, unmark ready, advance head.

### 3. Runtime integration
- `IpcManager::GetTls`: lazily create a per-thread SHM server+client named
  `clio-[tid]`; TID carried in `SaveTaskArchive`.
- `Worker::Worker()` calls `GetTls`; `Worker::Run` does `shm_server->Recv` with
  `DONTWAIT` instead of polling FutureShm completion.
- `Admin::ClientConnect`: return a vector of worker PIDs (not the SHM queue
  offset).
- `ipc_cpu2cpu` SendIn/RecvIn: propagate return TID; establish client connection
[...]

PR #789, opened 2026-07-22
docs(sched): plan for migrating blocked/in-flight tasks off a stalled worker (#785)

### LINKED ISSUE #785: Work stealing and completion signalling must cover blocked/in-flight tasks, not just lane arrivals
## Problem

Issue #781 stopped a bad (non-yielding / mislabeled) task from wedging the *runtime*: `RuntimeMapTask` now routes **new** work around a stalled worker, and the `WorkOrchestrator` monitor thread detects the stall. But everything that is **already in flight on that worker is still stranded**, so a single bad task can still stall an unbounded number of logical operations.

Four pieces of a worker's state are reachable only by that worker's own thread:

1. `Worker::assigned_lane_` — MPSC ring; the wedged worker is its only consumer, so the queued backlog never runs. (#781 left "steal-safe MPSC pop" as an open question.)
2. `blocked_queues_[4]`, `periodic_queues_[4]`, `retry_queue_` — plain worker-private `std::queue`, no lock, touched only by the owner thread. Not stealable at all today.
3. `Worker::event_queue_` — the MPSC ring of completed-subtask `Future`s. Producers push, but **only the wedged worker pops**, so a subtask that finished *elsewhere* still cannot wake its parent.
4. Tasks suspended on `co_await` live in **no queue at all** — they are reachable only via the subtask future's owning `parent_task_` pointer plus the raw `parent->EventQueue()` / `parent->Lane()` addresses cached in their `RunContext`.

The root design flaw is that **completion signalling is worker-addressed**: `IpcCpu2Self::SendOut` reads `parent_task->EventQueue()` and pushes into *that specific worker's* ring, then `AwakenWorker(parent_task->Lane())`. Consequently a parent task is permanently bound to the worker it first ran on. That binding is also what makes reassignment hard: moving a blocked task means rewriting signal targets that in-flight subtasks are concurrently reading. (The same worker-addressed design already produced the lost-wakeup class of bugs — see the `AwakenAllWorkers` fallbacks in `IpcManager::AwakenWorker`, #680.)

## Goal

Work stealing / reassignment must cover blocked and suspended tasks, not just fresh lane arrivals — and completion signalling must survive reassignment without a lost or duplicated wakeup.

## Scope

- Make completion signalling **late-bound**: a finishing subtask marks its parent runnable and hands it to the scheduler *at signal time*, instead of pushing into a worker's ring chosen when the parent first ran. Reassignment then needs no signal rewriting.
- Make the parked state (blocked / periodic / retry) stealable under a park lock that is never held across `ExecTask`.
- Rescue a stalled worker's lane backlog by **transferring the lane** to a freshly spawned worker (`SpawnAdditionalWorker` + `lane->SetTid`) rather than draining an MPSC ring from a foreign thread.
- Per-task scheduling state machine (CAS) so a steal can never double-execute or lose a task.
- Audit cross-thread resume: Boost fiber stack slab ownership, `CLIO_CUR_WORKER` caching across `co_await`, and per-thread `IpcManagerTls` (`clio-<pid>-<tid>`) affinity.
- Deterministic repro: extend `clio_run_thrpt_bench --test-case sched_variety` with parent/child dependency chains so an in-flight stall is reproducible, not incidental.

## Follow-up to

#781 (merged in #782). This picks up the two explicit TODOs left in `DefaultScheduler::LoadBalance()` / `StealWork()` and `WorkOrchestrator::SpawnAdditionalWorker()` / `RetireWorker()`.

lukemartinlogan:
Design document only — **no code yet**. Opening as a draft so the plan can be reviewed before implementation starts.

Closes #785 when the implementation lands; this PR is currently just `project_blocked_task_stealing.md`.

## Why

#781 (merged in #782) stopped a mislabeled / non-yielding task from wedging the runtime, but only for **new** work: `RuntimeMapTask` routes around a stalled worker and the monitor thread detects the stall. Everything already committed to that worker is still stranded, so one bad task can still stall an unbounded dependency tree. This picks up the TODOs #781 left in `LoadBalance()` / `StealWork()` / `SpawnAdditionalWorker()`.

## Root cause

Completion signalling is **worker-addressed**. `IpcCpu2Self::SendOut` pushes into `parent_task->EventQueue()` and wakes `parent_task->Lane()` — raw addresses stamped into the parent's `RunContext` the first time it was popped off a lane (`worker.cc:429-432`) and never revisited. A parent is therefore pinned to its first worker for life, and a subtask that completed on a *healthy* worker still cannot wake it. The same design already produced the #680 lost-wakeup class.

## Design (as decided during review)

* `LoadBalance()` is the **only** migrator — single writer, so no CAS state machine is needed.
* Event queue gets a consumer mutex. `ctp::ipc::mpmc_ring_buffer` already *is* `mpsc + RING_BUFFER_LOCK_POP`, so this is a one-line type change.
* A per-task `sig_mtx_`/`sig_gen_` handshake, because the consumer mutex alone still loses wakeups (a producer reads the queue pointer unsynchronised and can push into the orphaned queue after the drain).
* `co_await`-suspended tasks get parked in `blocked_queues_` so they are enumerable at all.
* The lane is **transferred**, not drained — sidesteps #781's open "steal-safe MPSC pop" question and keeps the SHM ring lock-free.
* Net-worker roles are deleted in favour of `TaskGroup` affinity: three groups following socket ownership (peer-send, peer-recv, client), each bound to a worker of its own.

## Findings worth reviewing even if the plan changes

Several of these are pre-existing issues found while verifying the design:

1. **Live bug on dev.** `RuntimeMapTask` consults the task-group map *before* the periodic role table, and the insert is first-writer-wins. `SendTask` and `RecvTask` are both in group 0, so whichever routes first binds the group and the other follows it there, bypassing its role. The documented send/recv split — keep outbound DEALER sends off the recv worker so inbound SWIM probes still get polled — is already not happening.
2. **`blocked_queues_` is vestigial.** Nothing reaches `AddToBlockedQueue`'s `YieldTimeUs() == 0` branch; `ProcessBlockedQueue`'s own re-add at `worker.cc:951` is dead code behind a `continue`. The four buckets and `WorkerStats::num_blocked_tasks_` have never been exercised.
3. **`ProcessBlockedQueue` resumes unconditionally** — no readiness check. Harmless today because the queue is empty; #705-class corruption the moment event-waiting tasks are parked there.
4. **`ReschedulePeriodicTask` silently drops** a periodic task whose `Lane()` is null (`worker.cc:1247-1251`) — no log, task gone. Worth fixing regardless.
5. **`task_group_map_` is insert-if-absent and never updated**, and the group early-return precedes every health check. Migrating a group off a stalled worker sends it straight back on the next period. The map also holds raw `Worker*`, which dangles once `RetireWorker` exists.
6. **Published worker tids go stale under an elastic pool.** `ClientConnect` publishes worker OS tids at connect time (#642) so SHM clients can address `clio-<pid>-<tid>` mailboxes; a client holding a *retired* worker's tid addresses a dead mailbox. Constrains `RetireWorker` independently of this work.

## Phases

`P0` repro + telemetry → `P1` replacement workers + lane transfer → `P2a` park `co_await` tasks + readiness scan → `P2b` dissolve net roles into task groups → `P2c` periodic migration → `P2d` event queue + blocked migration → `P3` observability.

Each is independently mergeable. P2b fixes finding 1 and is independently valuable.

## Not in scope

Idle-thief work stealing (`StealWork()` stays a stub), preemption of a running task, cross-node stealing, and curing a task wedged *inside its own handler* — this contains a stall, it does not cure one.

## Open questions

Five, listed in §8 of the doc, none blocking: awaited-future handle (raw `void*` vs owning `Future`), readiness-scan backoff keying, wedged-worker readmission, all-workers-stalled behaviour, and whether the four periodic buckets survive.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -148,3 +148,4 @@ claude-log.json
 # Windows firewall allow-rule marker (scripts/windows/allow_firewall.ps1)
 .clio_firewall_allowed
 context-runtime/test/integration/collective_restart/docker-compose.override.yml
+build-tsan/
--- a/context-runtime/include/clio_runtime/config_manager.h
+++ b/context-runtime/include/clio_runtime/config_manager.h
@@ -164,8 +164,16 @@ class ConfigManager : public ctp::BaseConfig {
    * spawn budget in WorkOrchestrator (one replacement per core).
    */
   u32 GetElasticLaneHeadroom() const {
+    // Bounded deliberately. Each reserved lane costs queue_depth x priorities
+    // slots in the queue segment, so scaling this with core count would add
+    // ~128 lanes of segment on a large machine purely to hold replacements
+    // that may never be spawned. Rescues are rare and replacements are
+    // recycled, so a modest fixed ceiling is enough: the pool only ever needs
+    // as many replacements as there are workers wedged AT ONCE.
+    static constexpr u32 kMaxElasticLanes = 8;
     int ncpu = ctp::SystemInfo::GetCpuCount();
-    return (ncpu > 0) ? static_cast<u32>(ncpu) : 8;
+    u32 want = (ncpu > 0) ? static_cast<u32>(ncpu) : 8;
+    return want < kMaxElasticLanes ? want : kMaxElasticLanes;
   }
 
   /**
--- a/context-runtime/include/clio_runtime/config_manager.h
+++ b/context-runtime/include/clio_runtime/config_manager.h
@@ -34,6 +34,7 @@
 #ifndef CLIO_RUNTIME_INCLUDE_MANAGERS_CONFIG_MANAGER_H_
 #define CLIO_RUNTIME_INCLUDE_MANAGERS_CONFIG_MANAGER_H_
 
+#include "clio_ctp/introspect/system_info.h"
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -155,6 +156,17 @@ class ConfigManager : public ctp::BaseConfig {
    * @return Number of worker threads for task execution
    */
   u32 GetNumThreads() const { return num_threads_; }
+
+  /**
+   * issue #785: extra task lanes reserved for elastic replacement workers.
+   * Lanes are indexed by worker id, so replacements spawned by a stall rescue
+   * need lanes allocated up front or they can never be routed to. Matches the
+   * spawn budget in WorkOrchestrator (one replacement per core).
+   */
+  u32 GetElasticLaneHeadroom() const {
+    int ncpu = ctp::SystemInfo::GetCpuCount();
+    return (ncpu > 0) ? static_cast<u32>(ncpu) : 8;
+  }
 
   /**
    * Get task queue depth per worker
--- a/context-runtime/include/clio_runtime/scheduler/default_sched.h
+++ b/context-runtime/include/clio_runtime/scheduler/default_sched.h
@@ -111,7 +111,8 @@ class DefaultScheduler : public Scheduler {
   std::vector<Worker *> elastic_workers_;
 
   /** #785 progress watchdog. Monitor-thread only, so plain members. */
-  static constexpr double kNoProgressAlarmSec = 10.0;
+  static constexpr double kNoProgressAlarmSec = 10.0;   ///< nothing executing
+  static constexpr double kAllBlockedAlarmSec = 60.0;   ///< all blocked (ambiguous)
   u64 last_progress_count_ = 0;
   double last_progress_us_ = 0.0;
   bool progress_alarm_raised_ = false;
--- a/context-runtime/include/clio_runtime/scheduler/default_sched.h
+++ b/context-runtime/include/clio_runtime/scheduler/default_sched.h
@@ -110,6 +110,11 @@ class DefaultScheduler : public Scheduler {
    *  fresh work, but still checked for stalls so cascades are rescued. */
   std::vector<Worker *> elastic_workers_;
 
+  /** #785: least-loaded live worker with a lane, excluding two given workers.
+   *  Used to re-place a stalled worker's queued backlog. Monitor thread only. */
+  Worker *PickLeastLoadedLive(double now_us, Worker *avoid_a,
+                              Worker *avoid_b) const;
+
   Worker *scheduler_worker_;              ///< Worker 0: metadata + small I/O
   std::vector<Worker *> io_workers_;      ///< Workers 1..N-3: large I/O
   Worker *net_send_worker_;               ///< Worker N-2: kSend / kClientSend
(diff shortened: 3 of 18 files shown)


lukemartinlogan:
## CI verification status (commit `c7f87a4`)

| Workflow | Result |
|---|---|
| CI Linux | ✅ success — includes **asan, ubsan, msan** |
| CI macOS | ✅ success |
| CI Adapters | ✅ success |
| CI VFD | ✅ success |
| Build Core Containers | ✅ success |
| cpplint | ✅ success |
| CI Windows | ❌ — attributed below |

**Windows failures are environmental, not code.** Three jobs failed; `build-test (windows-2022)`, `build-test (windows-2025)`, `icx (windows-2022)` and `boost (windows-2025)` all passed:

- `build-test (windows-11-arm)`: only `cte_bdev_leak_stress_force_net`, which also fails on `dev` (documented flake from #781: different test each run, clears on retry).
- `icx (windows-2025)` and `icx (windows-2025-vs2026)`: linker `Access is denied.` — runner file-lock transient, no compile or test error.

Different configs passing and failing between rounds is the signature of infrastructure flake rather than a regression.

### One real regression was caught here and fixed

Earlier in this branch, `cr_all_safe_bdev_tests` SEGFAULTed on **all three** Windows configs while five consecutive `dev` runs were clean. Root cause: the rescue handed the donor's event queue to the replacement while the donor kept the same pointer — **two consumers on a single-consumer MPSC ring**. Invisible on Linux, invisible under TSan, fatal on slower Windows runners. Fixed in `4b88cce`; that SEGFAULT is now gone everywhere.

Auditing for the same class found two more, both fixed before CI saw them:
- adopted event queues were never forwarded on a *cascading* rescue, stranding inherited queues
- `TaskLane::Push` is `WAIT_FOR_SPACE`, so pushing from the **monitor thread** into a full lane would spin forever — freezing stall detection, rescues and the watchdog

§9c of the design doc records the invariant this design creates and the review question it demands.

### Local verification

- stall-migration suite: **9/9, six consecutive runs**, 97–102s, zero false alarms
- regression sweep: 272 tests across 12 binaries, 0 failures
- TSan: clean in every changed file (remaining reports are pre-existing `container.h` / `ring_buffer.h`)

### Still open — reviewer input wanted

- **Deadlock is detected, not prevented.** A true dependency cycle is a property of what callers request, not of the scheduler. The watchdog reports it with per-worker state instead of hanging silently; it cannot make it impossible.
- **Net/GPU workers are never migrated** (D7), so a wedge there is still a node-level outage until the rest of D8 lands.
- **D8-1 (network task-group split) is unverified.** No test in any environment I have exercises it — the admin network periodics are never spawned by the co-located client-mode runtime the unit tests use. It needs a networked or multi-node run before merge.

PR #795, opened 2026-07-22
fix(organizer): treat a zero access timestamp as a sentinel, not a time (#792)

### LINKED ISSUE #792: Frecency organizer derives blob age from steady_clock epoch — score depends on machine uptime (flakes cte_data_organizer_all)
## Summary

`cte_data_organizer_all` fails intermittently — **6 times across 6 different branches** in the last 100 failed CI runs, once per run, on Windows/icx. The failing assertion is in `DataOrganizer - Cold Blob Demoted`:

```
REQUIRE failed at context-transfer-engine/test/unit/test_data_organizer.cc:441
Condition: score > 0.2f
[INFO] cold blob score after organizer rounds: 0.0900307
```

The spread across unrelated branches with no code in common points at an environmental dependency, and the arithmetic identifies it: **the frecency score depends on how long the CI runner has been booted.**

## The arithmetic

From `FrecencyDataOrganizer` (`frecency_organizer.h`):

```
kRecencyHalfLifeSec = 600.0   kFreqSaturation = 10.0   kRecencyWeight = 0.5
score = 0.5 * 2^(-age/600) + 0.5 * n/(n+10)
```

The test writes the blob once (`n = 1`) and sleeps `5 * 500ms = 2.5s`. Expected:

```
score = 0.5 * 2^(-2.5/600) + 0.5 * (1/11) = 0.4988 + 0.0455 ≈ 0.545
```

Observed `0.0900307`. Solving for age:

```
0.0900307 = 0.5 * 2^(-age/600) + 0.0455
2^(-age/600) = 0.08915  →  age ≈ 2093 s ≈ 35 minutes
```

So the organizer scored a blob written 2.5 seconds earlier as if it were **35 minutes old**.

## Root cause

`GetCurrentTimeNs()` (`core_tasks.h:77`) is:

```cpp
return static_cast<clio::run::u64>(
    std::chrono::steady_clock::now().time_since_epoch().count());
```

`steady_clock`'s epoch is arbitrary — on Linux and Windows it is effectively **system boot**. Meanwhile several `BlobInfo`/`TagInfo` constructors zero-initialize `last_modified_(0)` / `last_read_(0)`.

`ComputeScore` does:

```cpp
Timestamp last_access = std::max(stat.last_read_, stat.last_modified_);
if (now > last_access) age_sec = (now - last_access) / 1e9;
```

With `last_access == 0`, `age_sec` becomes `now / 1e9` — **the machine's uptime in seconds**. A timestamp meaning "not recorded" is silently read as "as old as this machine has been running."

That makes the test outcome a function of runner uptime:

| runner uptime at test time | score | result |
|---|---|---|
| < ~17 min | > 0.2 | passes |
| > ~17 min | ≤ 0.2 | **fails** |

~2093 s ≈ 35 min matches a Windows/icx job, which builds far longer before reaching ctest than the Linux jobs that never fail this test. That is exactly why the flake is platform-skewed without being platform-specific in mechanism.

## What I have and have not established

**Established:** the clock is `steady_clock`; the constructors zero-init; a zero timestamp yields `age == uptime`; the observed score corresponds to an age matching plausible runner uptime and to nothing else in the system.

**Not established:** which specific path scored a `BlobInfo` whose timestamps were still zero. `PutBlobImpl` *does* set `blob_info_ptr->last_modified_ = now` (`core_runtime.cc:1503`), so the likely candidate is a race — the organizer runs on a 500 ms period and iterates `tag_blob_name_to_info_`, so a round landing between map insertion and the timestamp write would score a zeroed record and persist the bogus score via `ReorganizeBlobInternal`. Worth confirming before fixing.

## Why this matters beyond the test

If a zeroed timestamp reads as "ancient", the organizer will demote freshly written blobs to the slow tier. The data organizer's entire purpose is inverted for any blob it happens to observe in that window — a correctness issue, not just a flaky assertion.

## Suggested fix

1. **Treat 0 as a sentinel, not a time.** In `ComputeScore`, if `last_access == 0`, skip the blob or treat recency as 1.0 (just-written) rather than computing an age from the epoch.
2. **Do not derive an absolute age from `steady_clock`'s epoch.** Either store a wall-clock stamp, or store the value only as a difference against a stamp taken from the same clock and never compare against an unset zero.
3. Close the insert/stamp race so a `BlobInfo` is never visible to the organizer before its timestamps are set.
4. The test's `REQUIRE(score > 0.2f)` should not be an absolute constant; once the above is fixed it is a fair assertion, but it currently encodes an assumption about uptime.

## Reproducing

Deterministic if you can control uptime: the test fails whenever the machine has been up longer than ~17 minutes when it runs, given a blob whose timestamps are observed as zero.

lukemartinlogan:
Fixes #792.

## The bug

`FrecencyDataOrganizer::ComputeScore` derives a blob's age from `now - max(last_read_, last_modified_)`. Those timestamps come from `GetCurrentTimeNs()`, which is `std::chrono::steady_clock` — **whose epoch is arbitrary**, and is effectively machine boot on Linux and Windows.

`BlobInfo` is constructed with those fields zeroed and stamped shortly after in `PutBlobImpl`, so a periodic organizer round can observe one in between. When it does, `now - 0` is not an age — it is the entire steady_clock counter, i.e. **the host's uptime**.

## Why it matters beyond a flaky test

The organizer demotes brand-new data to the slow tier, inverting the purpose of frecency scoring. Any blob it happens to catch in that window is treated as ancient.

It also made `cte_data_organizer_all` flaky in a way that looked inexplicable: the test failed on runners that had been up longer than ~17 minutes and passed otherwise, appearing 6 times across 6 unrelated branches with no code in common.

## The arithmetic

With `score = 0.5·2^(-age/600) + 0.5·n/(n+10)` and `n = 1`, the CI failure value of `0.0900307` solves to an age of **~2093 s (~35 min)** for a blob written **2.5 s** earlier. Uptime is the only quantity in the system on that scale.

## The fix

- `CollectOrganizerBlobStats` skips blobs whose timestamps are both still zero, alongside the existing empty-blob skip. This costs nothing — the next round 500 ms later rescores them correctly.
- `ComputeScore` treats `last_access == 0` as just-accessed rather than deriving an age from the epoch, as a second line of defence for any other caller.

A blob with no recorded access is new, not ancient.

## Verification

`test_data_organizer` 6/6 across two runs. Cold-blob score is now **0.5454**, matching the predicted `0.5·2^(-2.5/600) + 0.5/11 = 0.544`, versus `0.0900307` in the CI failures.

## Note on scope

This fixes the *interpretation* of a zero timestamp. It does not change `GetCurrentTimeNs()` itself — using a `steady_clock` epoch as though it were an absolute timestamp is a broader pattern worth auditing separately, since any other code comparing an unset zero against it has the same latent bug.

### DIFF
--- a/context-transfer-engine/core/src/core_runtime.cc
+++ b/context-transfer-engine/core/src/core_runtime.cc
@@ -1988,6 +1988,14 @@ void Runtime::CollectOrganizerBlobStats(clio::run::u32 replica_id,
         // organizer replica. The key hash partitions the blob space
         // disjointly across the `organizer_tasks` periodic replicas.
         if (blob_info.blocks_.empty()) return;
+        // Skip blobs whose access timestamps have not been written yet. A
+        // BlobInfo is constructed with last_modified_/last_read_ zeroed and
+        // stamped shortly after (PutBlobImpl), so a periodic organizer round
+        // can observe one in between. Scoring it would treat the zero as a
+        // real timestamp and derive an age from the steady_clock epoch — see
+        // the sentinel note in FrecencyDataOrganizer::ComputeScore (#792).
+        // Skipping costs nothing: the next round rescores it correctly.
+        if (blob_info.last_modified_ == 0 && blob_info.last_read_ == 0) return;
         if (num_replicas > 1 && (hasher(key) % num_replicas) != replica_id) {
           return;
         }
--- a/context-transfer-engine/core/src/data_organizer/frecency_organizer.cc
+++ b/context-transfer-engine/core/src/data_organizer/frecency_organizer.cc
@@ -45,7 +45,23 @@ float FrecencyDataOrganizer::ComputeScore(const OrganizerBlobStat &stat,
   // steady-clock ns; guard against a stamp taken after `now` was sampled.
   Timestamp last_access = std::max(stat.last_read_, stat.last_modified_);
   double age_sec = 0.0;
-  if (now > last_access) {
+  if (last_access == 0) {
+    // No access has been recorded yet — the blob was observed between its
+    // insertion into the metadata map and the write of its timestamps.
+    //
+    // Zero is a sentinel, NOT a point in time. Computing `now - 0` here
+    // would yield the full value of the steady_clock counter, whose epoch
+    // is machine boot, so a blob written seconds ago would score as though
+    // it were as old as the machine's uptime — demoting brand-new data to
+    // the slow tier, and making the outcome depend on how long the host had
+    // been running. That is issue #792; it flaked cte_data_organizer_all
+    // whenever CI runners had been up longer than ~17 minutes.
+    //
+    // Treat it as just-accessed: a blob with no recorded access is new, not
+    // ancient. Collection also skips these (see CollectOrganizerBlobStats),
+    // so this is a second line of defence for any other caller.
+    age_sec = 0.0;
+  } else if (now > last_access) {
     age_sec = static_cast<double>(now - last_access) / 1e9;
   }
 


lukemartinlogan:
## This fixes more than a flaky test: WAL-recovered blobs were being demoted after every restart

Audit follow-up that materially strengthens the case for this PR.

### WAL replay leaves every recovered blob with zeroed timestamps

`core_runtime.cc:3676`, replaying `TxnType::kCreateNewBlob`:

```cpp
BlobInfo blob_info;                      // default ctor
blob_info.blob_name_ = txn.blob_name_;
blob_info.score_     = txn.score_;
tag_blob_name_to_info_.insert_or_assign(composite_key, std::make_shared<BlobInfo>(blob_info));
```

The default constructor (`core_tasks.h:839`) zeroes `last_modified_`, `last_read_` **and** `access_count_`, and replay never stamps them. The WAL transaction structs (`transaction_log.h`) carry no timestamp fields at all, so there is nothing to restore them from.

So after a WAL recovery, *every* recovered blob has `last_modified_ == last_read_ == access_count_ == 0` — not for a brief window, but permanently.

### What the organizer did with that, before this PR

```
recency   = 2^(-uptime/600)     -> ~0 once the host has been up a while
frequency = 0 / (0 + 10)        =  0        (access_count_ is zeroed too)
score     = 0.5*recency + 0.5*frequency -> ~0.0
```

Every recovered blob scored at or near **0.0 — the minimum** — and was therefore demoted to the slowest tier on the first organizer round after restart, discarding the `score_` that replay had just restored from the WAL.

That is a production data-placement bug, not a test artifact: restart the runtime and your entire working set migrates to slow storage regardless of how hot it actually is.

With this PR, `CollectOrganizerBlobStats` skips blobs with no recorded access, so recovered blobs keep their replayed `score_` until they are genuinely read or written. That is the correct behaviour.

### The convention here is already established elsewhere

`Runtime::TemporalSearch` (`core_runtime.cc:5359`) does exactly what this PR now does:

```cpp
Timestamp ts = blob_info.last_modified_;
if (ts == 0) return;    // "never been written" -> excluded
```

with the comment *"last_modified_ == 0 means the blob has never been written and is excluded from all time-range queries."* So zero-as-sentinel is the existing convention in this file; the frecency organizer was the one place that treated it as a real point in time. This PR makes it consistent.

### Related gap, NOT fixed here

Because the WAL does not persist timestamps, recovered blobs are also permanently invisible to `TemporalSearch` — it excludes them as "never been written," which is wrong: they *were* written, we simply lost the timestamp. Filed separately; it needs either timestamp fields in the WAL records or a replay-time stamp, and that is a schema decision beyond this fix.

Issue #800, opened 2026-07-22
Windows link-lock flake reduced ~3x by Defender exclusions but not eliminated — still ~22% of Windows runs


lukemartinlogan:
## Summary

The Windows Defender exclusions added in #790 **substantially reduced but did not eliminate** the Windows link-lock flake. I previously reported that fix as verified working ("link-lock=0"); that was based on a single run and was wrong. Measured over 40 CI Windows runs / 280 jobs:

| | jobs | link-locks | rate/job | P(≥1 red job per 7-job run) |
|---|---|---|---|---|
| before exclusions | 196 | 21 | **10.7%** | **55%** |
| after all exclusions | 56 | 2 | **3.6%** | **22%** |

The reduction is real and statistically significant — under the null hypothesis that the exclusions changed nothing, seeing ≤2 locks in 84 jobs has `p = 0.0045`. But roughly **1 in 5 Windows runs still gets at least one link-lock**, which makes this the largest remaining source of red Windows builds.

## The residual failures are both MSVC link steps

Both post-fix occurrences are `cmake -E vs_link_exe`, not the vcpkg applocal path:

```
run 29946221870, dev, icx (windows-2022)
[232/304] Linking CXX executable bin\test_async_io_exec.exe
FAILED: [code=1] bin/test_async_io_exec.exe
cmake.exe -E vs_link_exe --msvc-ver=1944 ...
Access is denied.
```

```
run 29914513154, icx (windows-2025)
[291/304] Linking CXX executable bin\test_restart.exe
FAILED: [code=1] bin/test_restart.exe
cmake.exe -E vs_link_exe ...
The process cannot access the file because it is being used by another process.
```

Crucially, in the first case the exclusion step **ran and succeeded** in the same job:

```
Exclude workspace from Defender (link-flake mitigation)   success
Defender exclusions added for D:\a\clio-core\clio-core
```

So the workspace was excluded and the link still failed. Either the exclusion does not take effect for this access pattern, or **Defender is not the (only) holder of the file.**

Note a third occurrence — `build-test (windows-2025)`, run 29909583074, the `vcpkg z-applocal` MSB3075 case — is excluded from the post-fix numbers above because it ran at 09:50, before the vcpkg exclusions landed at 10:19.

## Why "add more exclusions" is not the obvious next step

[...]

PR #801, opened 2026-07-22
ci(windows): retry the build once on a transient link lock (#800)

lukemartinlogan:
Mitigates #800.

## Why

The Defender exclusions in #790 helped, but measurement over **40 CI Windows runs / 280 jobs** shows they did not eliminate the link-lock:

| | jobs | link-locks | rate/job | P(≥1 red job per 7-job run) |
|---|---|---|---|---|
| before exclusions | 196 | 21 | 10.7% | **55%** |
| after all exclusions | 56 | 2 | 3.6% | **22%** |

Real reduction (`p = 0.0045`), but ~1 in 5 Windows runs still gets a red job from this — the largest remaining source of red Windows builds.

I previously reported the Defender fix as verified working based on a single run. That was wrong, and #800 records the correction.

## Why a retry rather than more exclusions

Both residual failures are MSVC link steps:

```
[232/304] Linking CXX executable bin\test_async_io_exec.exe
cmake.exe -E vs_link_exe ...
Access is denied.
```

In that job the exclusion step **ran and succeeded** — `Defender exclusions added for D:\a\clio-core\clio-core` — and the link still failed. So either the exclusion doesn't cover this access pattern, or Defender isn't the holder.

Since the holder is unconfirmed, adding another exclusion is the same guess that already only half-worked. A retry works regardless of which process is at fault.

## Why this can't mask real breakage

The build is incremental. A genuine compile error fails identically on the second pass and the step still goes red — the only cost is one wasted attempt. A transient lock clears in seconds and the retry redoes just the failed link.

## Retries stay visible

Each retry emits a `::warning::` annotation, so it appears in the run summary. This keeps the flake **countable** rather than silently absorbed, and gives an ongoing signal for whether the root cause is getting better or worse.

That matters given what I found earlier in this investigation: `--repeat until-pass:3` on ctest and the `-E` exclusions in `ci-linux.yml` were both already hiding real problems. A mitigation that hides its own firing would repeat that mistake.

## Scope

Applies to all three Windows jobs (build-test, icx, boost). **This is a mitigation, not a fix** — #800 stays open for the root cause, and the diagnostic proposed there (one run with `Set-MpPreference -DisableRealtimeMonitoring $true`) still settles whether Defender is responsible.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -131,6 +131,15 @@ jobs:
       - name: Build
         shell: pwsh
         run: |
+          # GitHub's pwsh shell runs with $ErrorActionPreference = 'Stop'. In
+          # PowerShell 7.3+ that can also apply to NATIVE commands via
+          # $PSNativeCommandUseErrorActionPreference, in which case a non-zero
+          # exit from cmake throws and terminates the script — the retry below
+          # would never run and this step would behave exactly as it does today,
+          # silently. Force both off so a failed build is an exit code we can
+          # branch on rather than a terminating error.
+          $ErrorActionPreference = 'Continue'
+          $PSNativeCommandUseErrorActionPreference = $false
           cmake --build build --config Debug -j $env:NUMBER_OF_PROCESSORS
           if ($LASTEXITCODE -ne 0) {
             Write-Host "::warning::BUILD-RETRY: first build attempt failed (exit $LASTEXITCODE); retrying once — see issue #800"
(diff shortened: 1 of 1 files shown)


lukemartinlogan:
## A live case this would have recovered

PR #802 — a two-file test-logging change, branched from `dev` and therefore **without** this retry — just failed CI on exactly the flake this PR mitigates:

```
job: icx (windows-2025-vs2026)
step: Configure and build (icx)

[..] Linking CXX executable bin\zmq_transport_full_test.exe
The process cannot access the file because it is being used by another process.
```

Nothing in that PR touches the build, Windows, or the linker. It is the link-lock, firing on an unrelated change and turning a clean PR red. With this retry in place the second pass would have redone that one link and the job would have gone green.

That is the fourth observed occurrence since the Defender exclusions landed, consistent with the ~3.6%/job residual rate measured in #800 — and a concrete illustration of why the residual still matters: it is not that Windows builds are broken, it is that every unrelated PR has roughly a 1-in-5 chance of picking up a red job it had nothing to do with.

PR #806, opened 2026-07-22
ci(conda): use conda-forge only, drop the Anaconda defaults channel

lukemartinlogan:
An infrastructure flake that fails conda-based jobs before any build runs.

## Symptom

`build-test (macos-26)` on PR #805 failed at **Install dependencies (conda)** — nothing to do with the change under test:

```
CondaHTTPError: HTTP 403 Forbidden for url
  <https://repo.anaconda.com/pkgs/main/noarch/repodata.json>
- The channel requires authentication.
conda-build install failed (attempt 3/3)
```

## Cause

`repo.anaconda.com` now returns **403** for CI/commercial use of the Anaconda `defaults` channels (`pkgs/main`, `pkgs/r`) without a license. Any conda job that queries `defaults` can hit this intermittently.

`ci-deps.sh` already tried to work around it with `conda tos accept` for those channels — but that doesn't help. It only *prepended* conda-forge and left `defaults` in the channel list, so conda still fetched `pkgs/main` repodata and 403'd. The existing 3× create-retry loop couldn't recover either: every retry hit the same forbidden channel. And accepting a ToS is irrelevant when the IP is refused outright rather than prompted.

## Fix

Drop `defaults` entirely, pin conda-forge:

**`CI/ci-deps.sh`**
- clear seeded channels, add conda-forge, `channel_priority strict`, `default_channels []` so conda never falls back to defaults
- `conda create ... -c conda-forge --override-channels`
- removed the now-pointless `pkgs/main` + `pkgs/r` `tos accept` calls

**`.github/workflows/install-conda.yml`**
- `channels: conda-forge` (was `conda-forge,defaults`)

conda-forge is a strict superset of what this project pulls from defaults, so nothing needed is lost.

## Verification

`bash -n CI/ci-deps.sh` and YAML parse on `install-conda.yml` both pass. No functional `defaults` / `-c defaults` reference remains anywhere under `.github/workflows` or `CI/`.

I cannot run the macOS conda flow locally, so the proof is the next macOS CI run — but the mechanism is unambiguous: the 403 comes from a channel this change no longer contacts.

## Context

Found while checking why #805 (a heap-buffer-overflow fix) had a red macOS job. It was this flake, not the fix — the same shape as most of what this flake investigation has surfaced: a red job caused by something other than the PR's content.

### DIFF
--- a/.github/workflows/install-conda.yml
+++ b/.github/workflows/install-conda.yml
@@ -33,7 +33,11 @@ jobs:
         with:
           auto-update-conda: true
           python-version: '3.12'
-          channels: conda-forge,defaults
+          # conda-forge only. The Anaconda `defaults` channels (pkgs/main,
+          # pkgs/r behind repo.anaconda.com) now 403 for CI use without a
+          # license and were an intermittent dependency-install flake. See
+          # CI/ci-deps.sh for the same change and the full explanation.
+          channels: conda-forge
           channel-priority: strict
 
       - name: Install conda-build
--- a/CI/ci-deps.sh
+++ b/CI/ci-deps.sh
@@ -150,18 +150,35 @@ ensure_conda() {
 # Ensure conda is available
 ensure_conda
 
-# Accept Conda Terms of Service for Anaconda channels
-echo -e "${BLUE}Accepting Conda Terms of Service...${NC}"
-conda tos accept --override-channels --channel https://repo.anaconda.com/pkgs/main 2>/dev/null || true
-conda tos accept --override-channels --channel https://repo.anaconda.com/pkgs/r 2>/dev/null || true
-echo -e "${GREEN}Conda ToS accepted${NC}"
-echo ""
-
-# Configure conda channels (add conda-forge if not already present)
-echo -e "${BLUE}Configuring conda channels...${NC}"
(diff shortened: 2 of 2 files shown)


lukemartinlogan:
## Frequency calibration — this is rare today, not chronic

Before this reads as a widespread flake, I measured how often it actually fires. Across 28 completed macOS CI jobs in the recent sample, the HTTP 403 appeared **once** — the occurrence on #805 that prompted this PR. Zero in the other 27.

So the honest framing:

- This is **not** a high-frequency flake right now. My earlier "any conda job can hit this intermittently" is true in principle but overstated as a present-tense rate.
- It is a **latent liability**: `defaults` is a channel CI has no license for, and Anaconda has been tightening 403 enforcement over time. The rate is more likely to rise than fall, and when it fires it fails the whole job before any build.
- The fix is still correct and low-risk regardless of current frequency — conda-forge is a superset of what this project pulls, and removing an unlicensed channel we don't need is the right state independent of the flake rate.

Priority: lower than I implied. Worth merging as hardening, not as an urgent fire. I have adjusted expectations here rather than leave the PR description implying this is reddening PRs constantly — it is not; it fired once in the sample I checked.

PR #813, opened 2026-07-23
ci(gpu): run the CUDA compile-check on PRs, not only after merge

lukemartinlogan:
Closes the CI blind spot that let a dev-breaking GPU regression through review.

## The gap

`GPU Tests (CUDA)` triggered **only on push to `main`/`dev`** — never on pull requests. So a change that breaks the CUDA device compile is invisible during PR review and only surfaces *after* merge, on `dev`.

That's exactly how the `shm_metadata_cache.h` device-pass break got in:
- #788 called host-only IpcManager members (`GetMetadataAllocator`) in a header nvcc compiles for the **device** pass
- #788's PR CI never ran GPU, so review saw green
- the break appeared on `dev` only post-merge (now fixed by #812)

Right now, `dev`'s GPU job is red from that break, and it's the **only** failing thing across all of CI.

## The fix

Add a `pull_request` trigger with the **same path filter**, so any PR touching buildable source runs the same compile-only CUDA build and catches this class of break before merge. PRs that don't touch those paths still skip it (via the shared YAML anchor).

Plus a `concurrency` guard (`cancel-in-progress` per ref) — the CUDA toolchain install + conda build is ~15 min, so a new push to a PR branch cancels the previous GPU run instead of queuing another.

**No change to what the job builds — only when it runs.**

## Why this reduces flakes

Every GPU-breaking change currently lands on `dev` before anyone sees it, then shows as red on the branch until fixed. Validating on PRs moves that detection left, so `dev` stays green and PR authors get the signal directly. It's the structural counterpart to #812 (which fixes the specific break): #812 stops the bleeding, this stops the next one.

## Tradeoff for reviewers

This adds a ~15-min CUDA build to PRs that touch `context-*`/CMake/conda paths. That's real CI cost. The alternative — GPU validated only post-merge — is what just let a regression through. The path filter and concurrency guard keep the cost bounded.

### DIFF
--- a/.github/workflows/gpu-tests.yml
+++ b/.github/workflows/gpu-tests.yml
@@ -3,7 +3,7 @@ name: GPU Tests (CUDA)
 on:
   push:
     branches: [main, dev]
-    paths:
+    paths: &gpu_paths
       - 'context-runtime/**'
       - 'context-transfer-engine/**'
       - 'context-transport-primitives/**'
@@ -11,7 +11,24 @@ on:
       - 'CMakePresets.json'
       - 'installers/conda/**'
       - '.github/workflows/gpu-tests.yml'
+  # Also run on PRs touching the same paths. Previously this workflow fired
+  # ONLY on push to main/dev, so a change that broke the CUDA device compile
+  # was invisible in PR review and only surfaced after merge, on dev — which
+  # is exactly how the shm_metadata_cache.h device-pass break (#812, from
+  # #788) landed. Running the same compile-only build on PRs catches these
+  # before merge. Same path filter, so PRs that don't touch buildable source
+  # still skip it.
+  pull_request:
+    branches: [main, dev]
+    paths: *gpu_paths
   workflow_dispatch:
+
+# One in-flight run per ref; a new push to a PR branch cancels the previous
+# GPU build rather than piling them up (the CUDA toolchain install + conda
+# build is ~15 min).
+concurrency:
+  group: gpu-tests-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build-cuda-conda:


lukemartinlogan:
## Audited every workflow's triggers — GPU was the only one that warranted this

After finding the GPU push-only gap, I checked whether other workflows have the same review blind spot (run only on push, so a breaking change is invisible in PR review and only surfaces on dev post-merge).

**Trigger audit of all workflows:**

| workflow | trigger | verdict |
|---|---|---|
| CI Linux/Windows/macOS/Adapters/VFD, cpplint | push **+ pull_request** | already PR-validated ✓ |
| **GPU Tests (CUDA)** | push-only | **blind spot — this PR fixes it** |
| Cluster Tests | push-only (main/dev) | investigated — see below, NOT worth changing |
| Build/Publish Conda, Pip, Binary | push-only (main/tags) | correct — they publish artifacts, must not run on PRs |
| Spack Install / Spack | push-only (main) | main-only release validation; low PR value |

## Why GPU warranted the change but Cluster Tests does not

**GPU** — its compile coverage is *unique*: nothing else in CI compiles the CUDA device pass. A device-compile break (like the #812 one from #788) has no other job that would catch it in PR review. Plus a fresh, currently-red demonstrated break. Clear need. ~15 min.

**Cluster Tests** — I checked its one historical failure (Jul 2): a compile break in `test_fuse_copy_workspace.cc` (`'chi'` pre-rebrand namespace). But that file is built with `-DCLIO_CTE_ENABLE_FUSE_ADAPTER=ON` by **CI Adapters too** (ci-adapters.yml line 87), which *is* PR-validated. So its compile coverage is **redundant** with an existing PR gate — Cluster Tests being push-only did not create a unique gap for that break. Its genuinely-unique value is multi-node *runtime* behavior (distributed / failover / recovery / migration), and there is **no evidence** of a distributed-runtime break reaching dev. At 17 min of docker-multi-node per PR, adding it would be speculative cost with redundant compile coverage.

So the audit's conclusion is narrow and deliberate: **PR-validate GPU (unique coverage + demonstrated break), leave the rest.** I specifically did not reflexively add PR triggers to every push-only workflow — that would trade real CI cost for coverage that mostly already exists.

lukemartinlogan:
## Merge order matters: #812 before this PR

Heads-up for whoever merges these. This PR (add GPU compile-check on PRs) currently shows its own `Build CUDA conda package` check as **failing** — that is correct and expected: the trigger now runs GPU on the PR and catches the `shm_metadata_cache.h` device-pass break that is on dev, which #812 fixes and this PR does not contain.

Implication for merge order:

- **Merge #812 first** (the fix). Once it lands, dev's GPU build is green, and this PR's GPU check goes green too.
- If this PR merges **before** #812, the GPU-on-PR trigger goes live while dev still has the break — so every open PR touching `context-*`/CMake/conda paths will immediately show a red GPU check until #812 lands. That is the opposite of the intent.

So: #812 -> then #813. Or merge both together. Just not #813 alone first.

(This PR's failing GPU check is itself the proof the trigger works — it caught a real dev break on a PR, pre-merge, which is exactly what it is for.)

PR #827, opened 2026-07-25
gpu_vector span tiers (DRAM mirror + pinned VRAM cache) and batch-fault hardening

lukemartinlogan:
## Summary
- The gpu_vector now owns its out-of-core span tiers as first-class API: `BuildHostMirror` (pinned DRAM tier filled THROUGH CTE GetBlob, published in `DeviceViewBase::host_mirror`), `PinnedSpan` (VRAM span cache with permanent and two-touch admission policies), `StageTransient` (wrap-fenced transient ring), plus `CopySpan`/`CopyStream`/`RecordCopyEvent`/`SpanStats`. Callers orchestrate *which* spans a launch needs; the vector does everything else.
- These tiers are what carried the llama.cpp integration from 0.19 to 27 tok/s decode (141×) and 161 tok/s prefill for a 17 GB model on an 8 GB GPU — see the companion llama.cpp PR.
- Batch fault machinery over the flat single-blob layout (EvictRun/FaultRun/BatchSweep), hit-in-waiting resolve semantics, and the fault-path fixes (blob bare-stem naming, sync-fault wait, claim-before-drain eviction) that made on-device faulting correct under stress.

## Motivation
Model-weight inference through the vector required both correctness hardening of the fault path and a DRAM+VRAM tier architecture the measurements repeatedly proved necessary (cross-token reuse is impossible in VRAM alone at these working-set sizes; zero-copy mapped reads plateau ~6 GB/s vs ~10.6 GB/s DMA on gen4 x8).

## Test plan
- [x] `cte_gpu_vector_matrix` — 8 cases including the new tier-contract case (mirror byte-exact vs CTE-seeded blob; PinnedSpan two-touch decline/admit/hit asserted exactly; transient ring wrap with CopySpan fills)
- [x] `cte_gpu_vector_llama_geom_cuda`, `cte_gpu_vector_oversubscribe_cuda` — updated to bare-stem blob seeding (both were silently reading zeros), now pass
- [x] `cte_gpu_vector_cuda`, `cte_gpu_vector_gguf_map_cuda` pass
- [x] End-to-end llama.cpp integration verified bit-exact on Qwen3-30B (OOC, compressed and uncompressed) and gemma (resident, compressed and uncompressed)

## Breaking changes
None. `DeviceViewBase` gains a `host_mirror` field (recompile required); all new Vector methods are additive with fault-path fallback on any failure.


lukemartinlogan:
Pushed `f6cc6336` onto this branch — four defects that made the GPU vector's compressed and oversubscribed paths silently wrong or unbuildable after the compressor became a core interposer:

1. **Compression on the vector path was a no-op that reported success.** The interposer's put gate only fires on `Context::compress_lib_`, and it never applied the `CLIO_CTE_COMPRESS_LIB` operator pin — that pin lives in `Compress()`/`DynamicSchedule()`, which the interposer bypasses by calling `CompressIntoShm()` directly. The `gpu_vector` builds its page tasks with a default `Context`, so every page was stored **RAW**, including the llama.cpp compressed weight store. Now 8.4 MB -> 109 KB (76.8x) with a clean host round-trip.
2. **`CompressIntoShm()` omitted the header's `compressed_size_`**, so every interposer-written blob looked "legacy" to the reader and fell back to the over-allocated fetch length as the frame size — exactly the case that field exists to fix, breaking zstd/lz4 reads.
3. **`Client::TryReadBlobShm()` host-`memcpy`d into a device pointer.** The `gpu_vector` faults pages straight into an HBM slot, so dev's default-on zero-IPC read segfaulted. This is what made `cte_gpu_vector_oversubscribe_cuda` crash; it passes now.
4. **cuSZ header leak.** `<cusz.h>` declares a global `typedef uint8_t byte_t` and reaches the kvhdf5 e2e TUs via `compress_factory.h`, colliding with `using kvhdf5::byte_t` (`cuda::std::byte`). Renamed to `kv_byte_t`.

Plus build gating: `clio_model_weights_roundtrip` is gated on the CAE client target existing (CAE is OFF in the CUDA presets), and `llm-debug` is now a preset that can actually build the vector weight path.

Still open and **pre-existing**: `cte_gpu_vector_compress_cuda` / `..._compress_fault_cuda` hang on the GPU demand-fault decompress. `compress_cuda` timed out before any of these changes; fixes (1) and (2) move the other two from passing vacuously to reaching that same real hang.

Paired llama.cpp change: lukemartinlogan/llama.cpp#1 (quantized + bigger-than-VRAM weight streaming). It depends on fix (3) here.

lukemartinlogan:
Pushed `4e7715f7` — family-policy blob naming (`gpu_family_idx_` + `FamilyOf` policy + FamFn lambda on `dev::vector`), compose-file codec pin for the compressor, assimilator `pool=` routing, and the fault-service `cudaMalloc` liveness fix. This closes the long-standing `cte_gpu_vector_compress_fault` / `huge_compress` / `compress` hangs — all three pass (128/128 on-device faults, 76.8x, bit-exact). New `cte_gpu_vector_llama_geom_cuda` covers the multi-block × external-writer × cold-fault gap that hid the family deadlock. Known pre-existing failures unchanged: `capacity`/`transaction`/`prefetch` under the cuszp env pin.

Paired llama.cpp change: lukemartinlogan/llama.cpp#1 (`40873f76b`).

PR #842, opened 2026-07-28
Default the main segment reservation to system RAM capacity

### LINKED ISSUE #727: the 1 GiB main shared-memory segment is hardcoded — make it configurable
Measured while profiling clio-agent's default-backend footprint (iowarp/clio-agent#893): the runtime daemon's main shm segment ('Initializing main shared memory segment: 1073741824 bytes') is a hardcoded default — not settable via clio.yaml / cte yaml / env (verified against installed dev wheels; only bdev capacity / tier capacity_limit are tunable, and those bound the data tiers, not this segment). It is lazily committed on the first data op (~1 GiB commit-charge, mostly non-resident on Windows), so it dominates the daemon's COMMITTED memory (~1.18 GiB under any write load) regardless of actual data volume. Ask: expose the segment size in the compose/runtime config so embedded/desktop deployments can bound commit charge; document the default.

lukemartinlogan:
## Summary
- The auto default for the main task-data segment is now `SystemInfo::GetRamCapacity()` (1 GiB fallback if introspection fails), mirroring the metadata segment: the reservation is a sparse memfd on Linux, so it costs only virtual address space, and it cannot be grown later without invalidating every client's mapping.
- Creation-time safety is unchanged in spirit: the cgroup-aware half-budget clamp now fires by design for the auto value (segment lands at half the process's memory budget), and a new off-Linux 1 GiB cap prevents the RAM-sized default from becoming a real file (macOS/BSD) or up-front commit charge (Windows) — the exact footprint problem #727 reported.
- Explicit configuration (`runtime: main_segment_size` yaml / `CLIO_MAIN_SEGMENT_SIZE` env) is untouched and still wins.

## Motivation
Closes #727. The configurability half landed earlier; this completes the issue by replacing the arbitrary quarter-budget/1 GiB auto default with a host-scaled reservation, consistent with the metadata segment.

## Test plan
- [x] `cr_ipc_internals_tests` passes (the #727 regression suite, updated for the new auto default)
- [x] `cr_autogen_coverage_tests` and `cr_autogen_coverage_force_net` pass (broke on a prior iteration of #727)
- [x] Daemon verified on a 12 GB host: requests 12,543,832,064 B (= RAM), clamps to 6,271,916,032 B, initializes sparse

## Breaking changes
None. Linux daemons see a larger (virtual) reservation clamped to half the memory budget; off-Linux keeps the historical 1 GiB.

### DIFF
--- a/context-runtime/include/clio_runtime/config_manager.h
+++ b/context-runtime/include/clio_runtime/config_manager.h
@@ -204,8 +204,9 @@ class ConfigManager : public ctp::BaseConfig {
    * Size of the main task-data segment (issue #727).
    * @return The explicit size when main_segment_size_ > 0 (yaml
    *         `runtime: main_segment_size` or CLIO_MAIN_SEGMENT_SIZE), else the
-   *         auto default: a quarter of the process's cgroup-aware memory
-   *         budget, capped at 1 GiB and floored at 64 MiB. Never 0.
+   *         auto default: the machine's RAM capacity (1 GiB if introspection
+   *         fails). The reservation is sparse; creation-time clamps in
+   *         ipc_manager.cc bound what the segment actually gets. Never 0.
    */
   size_t CalculateMainSegmentSize() const;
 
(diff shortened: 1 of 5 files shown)


lukemartinlogan:
First CI round failed the daemon-lifecycle trio (`cr_cli_runtime_command_tests`, `cr_cli_cte_wal_restart`, `cr_cli_compose_restart`) on docker deps-cpu and arm — a real regression from this PR, now fixed in 69dcee06:

**Root cause:** `clio_run stop` ends in `std::abort()`, and the SIGABRT core dump *materializes the entire sparse reservation as zeros* and streams it through the piped `core_pattern` handler (piped patterns bypass `RLIMIT_CORE=0`). dev already dumps the ~8 GB (RAM-sized) metadata reservation on every stop; making the main segment RAM-sized too doubled that to ~16 GB, and on the slower runners the daemon outlived the tests' 60 s shutdown wait. Measured locally: abort-to-death 65 s → 2.1 s.

**Fix:** `madvise(MADV_DONTDUMP)` on shared mappings ≥ 1 GiB in `SystemInfo::MapSharedMemory` — covers server create and client attach in one place. As a side effect this should also make every daemon stop *faster than dev baseline*, since the metadata reservation is no longer dumped either. Segments < 1 GiB remain in crash dumps.

(The windows-2025 `cte_tag_force_net` timeout is the known pre-existing force_net flake, unrelated to this PR.)

Issue #820, opened 2026-07-24
Worker-local task batching: coalesce same-blob PutBlob/GetBlob to kill the #680 write-token tail


lukemartinlogan:
## Problem

Positioned writes/reads that land on the **same blob** serialize hard. The clio-fs page model stores each 1 MiB file page as one blob, so a sequential 4 KiB workload puts **256 tasks on each page-blob**. Every PutBlob to a blob must hold that blob's #680 write token across the *entire* operation — allocation **and** the bdev data copy — and a contender re-polls on a periodic cadence (`kBlobWriteLockPollUs`, `core_runtime.cc:1554`).

Measured (QD1, RAM target, 4 KiB): p50 write **7.3 µs**, but mean **~160 µs** → ~6.3K IOPS, with a tail of **p99.9 ≈ 10 ms and max ~1 s**. Block-size sweep confirms the cause is writes-per-blob: max stall **1082 ms** (4k, 256/page) → 10.5 ms (64k, 16/page) → 8 ms (1M, 1/page).

The 256 writes to a page are to **disjoint byte ranges** — they don't actually conflict — but the token is per-blob, so they drain single-file.

### A. Vectored PutBlob / GetBlob (new first-class task shape)

Give `PutBlob`/`GetBlob` a **scatter-gather segment list** instead of a single `(offset, ShmPtr, size)`:

```
segments_ : vector<{ offset, ShmPtr data, size }>
```

The runtime acquires the blob write token **once**, extends/resizes to cover the union of segment ranges, and applies every segment in one pass (one metadata mutation, N region copies) — then releases. GetBlob reads each segment's range directly into that segment's `ShmPtr`. Segments are applied in list order, so overlaps resolve **last-writer-wins**, which is also the ordering the #680 token race currently gets *wrong*.

This is useful on its own as a client API (a scatter-gather write/read in one task), and it is the representation the batching layer below produces — so a coalesced batch needs **no contiguous gather buffer and no post-hoc scatter copy**: each parent's `ShmPtr` becomes a segment, and for GetBlob the runtime writes straight into each parent's buffer.

Touches: `PutBlobTask`/`GetBlobTask` (`core_tasks.h`) gain the segment vector (single-region ctor stays, as a 1-segment case); `PutBlobImpl`/`GetBlobImpl` (`core_runtime.cc`) loop segments under one token; SaveTask/LoadTask serialize the vector.

### B. Worker-local task batching (coalesce into vectored tasks)

[...]

Issue #841, opened 2026-07-28
RouteTask drops retryable tasks when submitted from a non-worker thread (CI ServerInit wedge)


lukemartinlogan:
## Description

When `RouteTask` gets `RouteResult::Retry`/`Dne` back from `RouteLocal` (container plugged or not yet registered), it re-queues the task on the **current worker's** retry queue:

```cpp
// context-runtime/src/ipc_manager.cc (~L3827)
if (result == RouteResult::Retry || result == RouteResult::Dne) {
  Worker *worker = CLIO_CUR_WORKER;
  ...
  if (worker) {
    worker->AddToRetryQueue(task_ptr);
  }   // <-- else: task silently dropped
}
```

On a **non-worker thread** — e.g. the main thread running `ServerInit`'s compose step — `CLIO_CUR_WORKER` is null, so the task is silently dropped. Any caller awaiting that task's Future hangs forever.

## Evidence (CI, `build-test (unit amd64)`, run 30323737127)

The job was cancelled at the 2h workflow timeout after 14 tests each burned a 300–600s ctest timeout (`cte_query_*`, `cte_tag_*`, `cte_core_performance`, `cte_*_force_net`, `cr_streaming_*`, ...). Each hung test shows the same signature:

- ServerInit compose logs `Compose: Creating pool ram::chi_default_bdev`, then `RouteLocal: Container not found for pool=PoolId(major:1, minor:0) container_id=4294967295 method=10` with **`worker=-1`** → the CreatePool task is dropped → ServerInit waits forever.
- The scheduler heartbeats for the full 300 s with `exec-time bins ... all zero` — the runtime is alive but never executes a single task.

Comparison with the last green dev run (30138267947, Jul 25): it contains **19,101** `RouteLocal returned 4` errors, all on `worker=0`/`worker=1` — worker threads re-queue and recover, and every test passes. The failing run has **108 `worker=-1` occurrences** — exactly the dropped-on-the-floor case. Transient container-not-found during init is common on slow 2-vCPU coverage runners; the drop turns it into a permanent hang.

## Expected vs Actual

- **Expected:** a task that fails to route transiently is retried regardless of which thread submitted it.
- **Actual:** submissions from non-worker threads are dropped on transient routing failure; the waiter hangs until the ctest/job timeout.

## Proposed fix

[...]

PR #843, opened 2026-07-28
fix(ipc): non-worker-thread route failures were silently dropped — startup wedge behind the unit-amd64 CI timeouts

lukemartinlogan:
## Symptom

`build-test (unit amd64)` fails at the 2 h job limit ([run 30323741257](https://github.com/iowarp/clio-core/actions/runs/30323741257/job/90164732419)): **14 tests time out** (cte_tag_construction, cte_tag_large, cte_query_* ×4, cr_streaming_* ×2, cr_ipc_internals, cte_core_workflow, cte_runtime_coverage_all, cte_parallel_overlap_stress, cte_bdev_fragmentation, cte_evict), burning ~80 min, and ctest never reaches the tail of the suite. All 14 pass locally in seconds. The gate has not been green since the #833 batch landed (every run since was cancelled or timed out).

## Diagnosis (from the CI log)

The hung tests share one signature — the daemon wedges **at startup**, mid-compose:

```
ServerInit  Compose: Creating pool ram::chi_default_bdev (module: clio_bdev, ...)
RouteLocal  Container not found for pool=PoolId(1,0) container_id=4294967295 method=10
RouteTask   RouteLocal returned 4 for pool=PoolId(1,0) method=10, worker=-1
```

…followed by 300 s of nothing but the scheduler's idle heartbeat, with **cumulative exec-time bins all zero** — the workers never executed a single task while the test's client waited forever.

`worker=-1` is the tell: compose submits `GetOrCreatePool` from the runtime **main thread** (`IpcCpu2Self::SendIn` → `RouteTask(force_enqueue=true)`). When the route hits a transient `Dne` (the admin pool's static container isn't resolvable yet — a startup race that 2-vCPU coverage-instrumented runners lose far more often than dev boxes), the handler did:

```cpp
if (worker) { worker->AddToRetryQueue(task_ptr); }   // CLIO_CUR_WORKER == null here
```

No `else`. The task is silently dropped, nobody ever completes its future, and ServerInit hangs — same defect class as the #785 periodic-task drop ("this used to return silently, DROPPING the periodic task").

In passing runs the same transient shows up on other methods (14/21/19/28/31 during admin-pool create) but those are periodic tasks whose next period papers over the loss. `GetOrCreatePool` gets no second chance.

## Fix

When the routing thread is not a worker, park the task on a **published** worker's retry queue (`CLIO_WORK_ORCHESTRATOR->GetWorker(0)`, bounded by `num_workers_published_`) instead of dropping it. Workers poll `ProcessRetryQueue()` every loop iteration and never sleep past `max_sleep` (50 ms), so no explicit wake is needed; the retry clears routed state and re-resolves, which converges once the container is registered. If no worker exists at all (pre-`SpawnWorkerThreads`), log at **Fatal** so the loss is loud instead of a silent hang.

## Not addressed here (deliberately)

The gate also runs the whole suite at `CLIO_CTP_LOG_LEVEL=0` (the `debug` preset), and the #820/#833 batch added per-task-pop DEBUG logging — `cte_bdev_fragmentation` emitted 3.8M log lines while *demonstrably progressing* through its 300 s window (slow-not-stuck, the ci-adapters e93def6d disease). If the gate is still tight after this fix, building the coverage phases at Info is the follow-up lever; kept out of this PR to keep the correctness fix isolated.

## Testing

- Local (Release, x86): full `test_tag_operations` battery, `test_query`, `test_vectored_blob_io` (4), `test_core_functionality` SHM-cache suite (5), `test_cte_shm_metadata_cache` (9) — all green with the fix.
- The startup race itself is timing-dependent (never reproduced in 8 pinned-2-CPU local runs); the CI gate on this PR is the real regression test.

### DIFF
--- a/context-runtime/src/ipc_manager.cc
+++ b/context-runtime/src/ipc_manager.cc
@@ -3829,8 +3829,28 @@ RouteResult IpcManager::RouteTask(Future<Task> &future, bool force_enqueue) {
       HLOG(kError, "RouteTask: RouteLocal returned {} for pool={} method={}, worker={}",
            (int)result, task_ptr->pool_id_, task_ptr->method_,
            worker ? (int)worker->GetId() : -1);
+      if (!worker) {
+        // Routed from a NON-worker thread (the runtime main thread during
+        // ServerInit/compose, via IpcCpu2Self::SendIn). This used to be a
+        // silent DROP: the transient Dne/Retry (e.g. the admin pool's static
+        // container not yet resolvable while compose submits GetOrCreatePool)
+        // lost the task forever, the submitter waited on a future nobody
+        // completes, and the runtime wedged at startup with provably idle
+        // workers — the CI unit-gate 300 s timeouts on cte_tag_construction /
+        // cte_query_* et al. Park it on a published worker's retry queue
+        // instead; every worker polls its retry queue each loop iteration and
+        // sleeps at most max_sleep, so the re-route needs no explicit wake.
+        worker = CLIO_WORK_ORCHESTRATOR->GetWorker(0);
+      }
       if (worker) {
         worker->AddToRetryQueue(task_ptr);
+      } else {
+        // No worker published yet (pre-SpawnWorkerThreads). Nothing can ever
+        // retry this task; make the loss loud instead of a silent hang.
+        HLOG(kFatal,
+             "RouteTask: dropping unroutable task (pool={} method={}) — no "
+             "worker exists to retry it",
+             task_ptr->pool_id_, task_ptr->method_);
       }
     }
     return result;


lukemartinlogan:
Superseded by #844 (issue #841), which contains this same worker-0 fallback + Fatal log but adds one missing piece: **binding `RunFuture` before parking**.

The compose path (`IpcCpu2Self::SendIn`, non-worker branch) calls `RouteTask` without ever binding `task_ptr->RunFuture()` — that normally happens later in `ProcessNewTask`/`RouteGlobal`, which this failed-route task never reaches. `Worker::ProcessRetryQueue` re-routes via `CLIO_IPC->RouteTask(task_ptr->RunFuture(), true)`, so with this PR's version the retried task hands `RouteTask` a default-constructed `Future` (null task pointer) — `future.GetTaskPtr()->pool_id_` then dereferences null on the retry tick, converting the 300s wedge into a daemon crash. Even absent a crash, the caller's `FutureShm` is owned by the `Future` that `SendIn` returns, so completion could never signal the waiter.

#844 parks the task the same way but first anchors the future exactly as `RouteManyToOne` does for its batch-parked members:

```cpp
task_ptr->RunFuture() = future;
task_ptr->RunFuture().GetTaskPtr() =
    clio::run::shared_ptr<Task>::WrapNonOwning(task_ptr.get());
```

Closing this one to keep a single CI signal; reopen if I've misread the retry path.

PR #844, opened 2026-07-28
Park unroutable tasks from non-worker threads instead of dropping them

### LINKED ISSUE #841: RouteTask drops retryable tasks when submitted from a non-worker thread (CI ServerInit wedge)

lukemartinlogan:
## Summary
- `RouteTask` silently dropped a task when `RouteLocal` returned `Retry`/`Dne` and `CLIO_CUR_WORKER` was null (non-worker thread, e.g. ServerInit's compose on the main thread) — the waiter then hung forever.
- Fix: park the task on worker 0's retry queue (mutex-guarded, safe cross-thread; the scheduler worker drains it every ≤32 loop iterations even when idle), and anchor the Future to the task's RunContext exactly as `RouteManyToOne` does so the retried task can still signal its waiter.

## Motivation
Fixes #841 — the cause of the `build-test (unit amd64)` 2-hour cancellation on recent runs (e.g. run 30323737127): ~14 unit tests each wedged in ServerInit for their full 300–600s ctest timeout. Log forensics: the last green dev run (30138267947) contains 19,101 transient `RouteLocal returned 4` failures, **all** on worker threads (re-queued, recovered); the failing run adds 108 with `worker=-1` — each a silently dropped task. Each hung test shows the identical signature: `Compose: Creating pool ram::chi_default_bdev` → `RouteLocal: Container not found ... worker=-1` → scheduler heartbeats for 300 s with all-zero exec bins (no task ever runs).

## Test plan
- [x] `cte_core_*` (15), `cte_query_*` (14), `cte_tag_*` (10 + force_net), `cr_all_safe_bdev_tests`, `cte_bdev_fragmentation`, `cte_functional_all` — all pass locally on the branch
- [ ] `build-test (unit amd64)` completes without the mass ctest timeouts (the race needs slow 2-vCPU coverage runners; `IsPlugged()` is hardwired false, so there is no deterministic hook for a unit repro — CI is the validation)

## Breaking changes
None.

### DIFF
--- a/CI/calculate_coverage.sh
+++ b/CI/calculate_coverage.sh
@@ -210,6 +210,12 @@ if [ "$DO_BUILD" = true ]; then
     # CLIO_CTE_ENABLE_COMPRESS=ON default leaked through and broke the
     # pkg_check_modules(liblz4 REQUIRED) call when lz4 wasn't in the
     # conda env.
+    # Info log level, not the debug preset's Debug (issue #845): Debug compiles
+    # every DEBUG HLOG into the hot path — one format+write syscall per task
+    # pop — which makes tests slow-not-stuck on the 2-vCPU coverage runners
+    # (cte_bdev_fragmentation alone emitted ~3.8M lines inside its 300 s ctest
+    # window). Same lever as e93def6d for the adapters gate; costs coverage of
+    # HLOG(kDebug) lines only.
     cmake --preset=debug \
         -DCLIO_CORE_ENABLE_COVERAGE=ON \
         -DCLIO_CORE_ENABLE_CONDA=ON \
@@ -217,6 +223,7 @@ if [ "$DO_BUILD" = true ]; then
         -DCLIO_CTE_ENABLE_ADIOS2_ADAPTER=OFF \
         -DCLIO_CTE_ENABLE_COMPRESS=OFF \
         -DCLIO_CORE_ENABLE_GRAY_SCOTT=OFF \
+        -DCLIO_CTP_LOG_LEVEL=1 \
         ${PHASE_CMAKE_ARGS}
 
     print_info "Building project..."
--- a/context-runtime/src/ipc_manager.cc
+++ b/context-runtime/src/ipc_manager.cc
@@ -3854,6 +3854,13 @@ RouteResult IpcManager::RouteTask(Future<Task> &future, bool force_enqueue) {
       }
       if (worker) {
         worker->AddToRetryQueue(task_ptr);
+      } else {
+        // No worker published yet (pre-SpawnWorkerThreads): nothing can ever
+        // retry this task — make the loss loud instead of a silent hang.
+        HLOG(kFatal,
+             "RouteTask: dropping unroutable task (pool={} method={}) — no "
+             "worker exists to retry it",
+             task_ptr->pool_id_, task_ptr->method_);
       }
     }
     return result;
--- a/context-runtime/src/ipc_manager.cc
+++ b/context-runtime/src/ipc_manager.cc
@@ -3862,9 +3862,21 @@ RouteResult IpcManager::RouteTask(Future<Task> &future, bool force_enqueue) {
         // hang ServerInit — a startup-timing window that CI's 2-vCPU runners
         // open reliably. Retry inline first: the window is transient and
         // normally closes within a few milliseconds.
+        //
+        // EXCEPT for periodic tasks. The admin container's Create spawns its
+        // periodic service tasks (kSend/kClientSend pumps, kHeartbeatProbe,
+        // kSystemMonitor) on THIS thread, and the container is only
+        // registered after Create returns — so for those, no amount of
+        // waiting here can succeed: the retry blocks the exact thread whose
+        // progress it awaits. Observed as 5 s x N stalls at every runtime
+        // init (unit-amd64 +19 min; the icx windows gate's rotating per-test
+        // timeouts, a retry storm filling the whole 300 s window). Fail them
+        // immediately instead — the same harmless outcome as the historical
+        // silent drop, but loud and with a completed future.
         constexpr int kInlineRetryMs = 5000;
-        for (int i = 0; i < kInlineRetryMs && (result == RouteResult::Retry ||
-                                               result == RouteResult::Dne);
+        for (int i = 0;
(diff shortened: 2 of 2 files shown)


lukemartinlogan:
Validation: this PR's `build-test (unit amd64)` **passed in 44m30s** — the same job was cancelled at the 2h limit on dev @ c7901dcf less than an hour earlier (run 30340001967: 100 `worker=-1` drops, same 300s wedge signature). With the drop fixed, zero ctest timeouts and the gate finishes with over an hour of headroom even while still building at Debug log level (#846 addresses that separately).

The two earlier Windows failures (`ctp_priv_umap_ll` on windows-11-arm, `cte_bdev_fragmentation` on icx/windows-2022) both passed on rerun — flakes. Remaining: asan + queued macOS jobs. `codecov/patch` is red because the added lines are the CI-race error path, which no test can deterministically execute (no plug/fault-injection hook; `IsPlugged()` is hardwired false).

lukemartinlogan:
Note on the codecov/patch 11% number: the uncovered lines are the fallback body itself, which executes only when the ServerInit routing race actually fires — nondeterministic by nature (it is what hung the CI gate, at maybe 1-in-20 runs on 2-vCPU runners). A deterministic unit test would need fault injection into GetStaticContainer to force the transient Dne from a non-worker thread; punting on that here to keep the fix minimal. The kFatal branch is deliberately unreachable in a healthy runtime. Empirical validation instead: unit amd64 44m30s / unit arm64 41m41s, versus the pre-fix 2h-timeout with 14 hung tests.

lukemartinlogan:
Merged `origin/dev` (post #840/#842). The conflict was semantic: dev's 868da55f landed its own fix for the same null-worker drop (inline retry 1ms×5s, then **fail** the future with rc=-1). Resolution combines both approaches:

1. **Inline retry first** (dev's validated fast path — the transient window normally closes in a few ms).
2. **On exhaustion, park instead of fail**: bind `RunFuture` (the RouteManyToOne pattern) and push onto worker 0's retry queue, so a longer-than-5s window still completes eventually instead of erroring ServerInit.
3. **Fail the future loudly only when no worker is published at all** (pre-SpawnWorkerThreads) — unblocking the waiter rather than the previous bare Fatal log.

Local on the merge: tag/query/vectored batteries green.

lukemartinlogan:
**Segfault root-caused and fixed** (91155efc). The merge resolution's parking fallback crashed `cte_tag_large` on icx windows-2025: a periodic admin task stayed unroutable for the full 5 s inline retry, got parked with `RunFuture`'s task pointer re-wrapped **non-owning** (the RouteManyToOne pattern), and worker 0's failed re-route re-queued it *through that non-owning handle* — when the popping iteration's owning reference died, the retry queue held a dangling pointer → use-after-free on the next drain ("parking task…" log line is followed by the segfault within ~0.5 s).

Lesson: a task still unroutable after 5000×1 ms is not the transient window this path exists for — parking it churns forever. Reverted to dev's 868da55f terminal-failure semantics verbatim, with the ownership hazard documented in place.

The PR's remaining delta vs dev is now: the #846 coverage-at-Info change + comments. Local tag battery green on the fix.

lukemartinlogan:
Root-caused the icx gate's rotating timeouts: they weren't runner flakiness after all. Dev's 5 s inline retry (868da55f) fires for the **periodic** admin service tasks spawned during the admin container's own Create — on the same thread that registers the container only after Create returns. The retry can never succeed there; it just stalls init 5 s per task (unit-amd64 44m → 1h04m; on the slowest Windows config the method=14/21/31 retry storms fill entire 300 s test windows — visible end-to-end in the cte_query_blob_multitag timeout log).

Fix (b53dd667): periodic tasks skip the inline retry and fail fast through the terminal path (same harmless outcome as the historical silent drop, but loud + completed future). One-shot tasks — the restart-replay compose the retry exists for — keep the full window.

Issue #823, opened 2026-07-25
Further optimization for GetBlob


lukemartinlogan:
We should add a private memory api for AsyncGetBlob. It takes as input a private memory raw pointer (char*). If the data is stored in the shared cache, we will copy from the shared cache. Otherwise, we will need to construct and submit a real AsyncGetblob task. There are two cases for the real task. 

1. If we are in runtime mode, we won't need to allocate a new FullPtr. We can just make our own with malloc allocator as the type. 
2. If we are in client mode, a shared-memory buffer will need to be allocated. The task should be marked as TASK_DATA_OWNER. when the user calls DelTask, the payload should be destroyed automatically.

We should verify all these cases. We should have a test where the runtime is started inline and when the runtime is started separate.

Issue #738, opened 2026-07-15
CTE: internal data organizer infrastructure (periodic DynamicReorganize + DataOrganizerFactory)


lukemartinlogan:
## Description
Data organization in the CTE is currently score-driven only at initial placement (PutBlob places blobs by tier score). After placement, nothing reorganizes data unless an external extension calls the RescoreBlob/ReorganizeBlob API. Prefetching/reorganization should be a first-class internal feature of the platform, not an extension the user must deploy.

## Proposed design
- **DynamicReorganize periodic RPC**: a new CTE method spawned periodically by the runtime. Its body is a one-liner delegating to the configured organizer: `organizer->Reorganize(this, task->replica_id_)`.
- **CTE config update**: new options:
  - `organizer: <name>` — which DataOrganizer to run (e.g. `frecency`; `none` disables)
  - `organizer_tasks: 1` — number of periodic task replicas to spawn to parallelize organization work
- **DataOrganizerFactory**: takes the organizer name, returns a `std::unique_ptr<DataOrganizer>`.
- **DataOrganizer**: abstract base with `Reorganize(Runtime *server, replica_id)`.
- **FrecencyDataOrganizer**: first concrete organizer. Rescores blobs by frecency (recently/frequently accessed blobs get higher scores) and applies the new scores through the server's ReorganizeBlob logic directly (no extra ReorganizeBlobTask spawned per blob).

## Acceptance Criteria
- [ ] Periodic DynamicReorganize task spawned from CTE Create when an organizer is configured
- [ ] organizer / organizer_tasks parsed from CTE YAML config
- [ ] FrecencyDataOrganizer rescoring based on access recency + frequency
- [ ] Unit test exercising the frecency organizer

PR #847, opened 2026-07-28
fix(cte): reorganize via ClearBlob under the blob write token (#753)

### LINKED ISSUE #753: CTE: concurrent GetBlob transiently fails while a blob is being reorganized (del-then-reput window)
## Description
`Runtime::ReorganizeBlobInternal` moves a blob between tiers by reading its data, **deleting** the blob (freeing capacity so the re-put needs only 1x space — see the Step 6.5 comment), then re-putting it at the new score. During the delete window the blob's metadata entry is gone, so a concurrent client `GetBlob` returns rc=1 (blob not found) and `Tag::GetBlob` throws.

This window predates #738 (an explicit `ReorganizeBlob` call always had it), but the periodic data organizer (#738/#744) now triggers moves in the background, so ANY client read can transiently fail whenever the organizer decides to move the blob being read. Observed in CI: `cte_data_organizer_all` on windows-2022 (Debug, slow) — the test's read loop overlapped the first organizer round and a mid-move read threw.

## Reproduction
Run a GetBlob loop against a blob whose frecency diverges from its current score while the organizer (organizer_period_ms small) is active on a slow/Debug build. windows-2022 CI reproduced it reliably on PR #744's first push.

## Possible fixes
- Allocate-before-free move (keep the BlobInfo entry and swap `blocks_` after writing the new placement) — removes the not-found window but needs transient 2x tier capacity, which the del-first ordering was specifically introduced to avoid (reorganize near-full tiers flaked 3/128).
- Keep the BlobInfo entry alive during the move (mark it 'moving', buffer the data server-side) and serve concurrent reads from the in-flight copy.
- Have GetBlob distinguish 'mid-reorganize' from 'truly absent' (e.g. a tombstone with the mover's token) and briefly retry.

## Acceptance Criteria
- [ ] A GetBlob issued while the organizer moves the same blob never fails with 'not found'
- [ ] Reorganize near-capacity still succeeds (no 2x-space regression, or an explicit fallback)

lukemartinlogan:
## Summary
- `ReorganizeBlobInternal` no longer moves a blob via `AsyncDelBlob` + `AsyncPutBlob`. It holds the blob's #680 per-blob write token for the whole move and uses `ClearBlob` + the internal helpers (`ReadData`/`ExtendBlob`/`ModifyExistingData`) inline — the `BlobInfo` entry never leaves the map, so a concurrent `GetBlob` can never see "blob not found" mid-move.
- The new placement is staged in a local `BlobInfo` and published in one co_await-free step; `GetBlobImpl` gains a torn-layout guard (range beyond current extent while the token is held → poll until the writer finishes) so readers never silently snapshot the emptied layout.
- The del-first capacity property is preserved: the old blocks are freed before the re-place, so the move still needs only 1x tier space (no 2x regression near-capacity).
- Side fixes that fall out of going inline: a writer landing between the read and the re-put can no longer be clobbered by the stale buffer; the move no longer bumps `last_read_`/`access_count_` (the old snapshot-and-restore stats dance is gone); `transform_flags_`/compression metadata survive the move (the del+reput path reset them); the WAL logs the new layout only after the publish, so a crash mid-move replays the old placement instead of losing the blob.
- New regression test: concurrent `GetBlob`s overlapping six tier-bouncing reorganizes must all succeed with intact data.

## Motivation
Fixes #753 — with the periodic data organizer (#738/#744) moving blobs in the background, ANY client read could transiently fail with "blob not found" whenever the organizer moved the blob being read (observed in `cte_data_organizer_all` on windows-2022 Debug CI).

Acceptance criteria from the issue:
- [x] A GetBlob issued while the organizer moves the same blob never fails with "not found" (entry never leaves the map; torn-window reads wait on the write token)
- [x] Reorganize near-capacity still succeeds (clear-before-re-place keeps the 1x-space property; placement failures retry, then restore at the original score into the just-freed capacity)

## Test plan
- [x] New regression case passes: `ctest -R cte_reorganize_all` (`ReorganizeBlob - Concurrent GetBlob during move`)
- [x] `cte_reorganize_all`, `cte_reorganize_force_net`, `cte_tag_reorganize`, `cte_data_organizer_all`, `cte_shm_metadata_cache_all` pass
- [x] Full `^cte_` suite (75 tests) matches baseline: the only failures are two pre-existing environment flakes (`cte_functional_force_net` SHM-cache case and `cte_putblob_priv_separate` rc-11 pool mismatch, #823) which reproduce identically with this change stashed
- [x] No new compiler warnings (clean build of `clio_cte_core_runtime`)

## Breaking changes
None. `ReorganizeBlob` rc semantics unchanged (rc=7 on a failed move, blob restored at its original score when possible).

### DIFF
--- a/context-transfer-engine/core/include/clio_cte/core/core_tasks.h
+++ b/context-transfer-engine/core/include/clio_cte/core/core_tasks.h
@@ -875,12 +875,15 @@ struct BlobInfo {
    * Source of placement generations: a runtime-wide monotonic counter, NOT a
    * per-blob increment.
    *
-   * Per-blob counting is unsafe across a re-create. ReorganizeBlob moves a blob
-   * by DelBlob + PutBlob, and the replacement BlobInfo starts at zero: a blob
-   * of the same size re-extends into the same number of blocks and lands on the
-   * SAME generation value it had before the move. A client that copied from the
+   * Per-blob counting is unsafe across a re-create (a DelBlob followed by a
+   * PutBlob of the same name): the replacement BlobInfo starts at zero, so a
+   * blob of the same size re-extends into the same number of blocks and lands
+   * on the SAME generation value it had before. A client that copied from the
    * old (now freed, possibly reallocated) blocks would then re-read an equal
    * generation and accept the bytes. A global counter cannot collide that way.
+   * (ReorganizeBlob used to be such a re-create; since issue #753 it keeps the
+   * BlobInfo alive and moves the blocks in place, but user-driven del+put
+   * re-creates remain.)
    */
   static std::atomic<clio::run::u64> &PlacementGenCounter() {
     static std::atomic<clio::run::u64> counter{1};
@@ -1066,6 +1069,17 @@ struct BlobInfo {
     ctp::ipc::atomic_ref<clio::run::u64> ref(write_owner_);
     clio::run::u64 expected = tok;
     ref.compare_exchange_strong(expected, 0);
+  }
+
+  /**
+   * True while some task holds the per-blob write lock. Advisory only — the
+   * answer can change the instant after it is read — so it is NOT a substitute
+   * for acquiring the lock. Readers use it to tell a torn mid-mutation layout
+   * (a reorganize has freed the old placement and not yet published the new
+   * one, issue #753) apart from a plain read past EOF.
+   */
+  bool IsWriteLocked() {
+    return ctp::ipc::atomic_ref<clio::run::u64>(write_owner_).load() != 0;
   }
 #endif  // CTP_IS_HOST
 };
--- a/context-transfer-engine/core/src/core_runtime.cc
+++ b/context-transfer-engine/core/src/core_runtime.cc
@@ -1994,6 +1994,20 @@ clio::run::TaskResume Runtime::GetBlobImpl(clio::run::shared_ptr<TaskT> &task) {
       CLIO_CO_RETURN;
     }
 
+    // Torn-layout guard (issue #753): a reorganize holds this blob's write
+    // token while it frees the old placement and publishes the new one; in
+    // that window blocks_ is transiently EMPTY, and snapshotting it would make
+    // ReadData silently succeed without reading a byte (it reports success for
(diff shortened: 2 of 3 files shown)


lukemartinlogan:
**fio-over-clio-fs A/B: no performance regression** (this branch vs its base c7901dcf, same box/build config, 4k psync via `LD_PRELOAD=libclio_cte_posix.so`, RAM tier, async cfs writes = default, medians of 3 iterations/side):

| workload | metric | baseline | #753 | delta |
|---|---|---|---|---|
| seqwrite nj=1 | IOPS | 13.3k | 18.6k | noise (ranges 11.4–19.3k vs 16.5–20.9k overlap) |
| seqwrite nj=8 | IOPS / mean lat | 8,391 / 950µs | 8,440 / 945µs | +0.6% / −0.5% |
| seqread nj=8 | IOPS / mean lat | 12,631 / 630µs | 13,800 / 576µs | +9.3% / −8.5% |
| randread nj=8 | IOPS / mean lat | 13,297 / 598µs | 13,358 / 596µs | +0.5% / −0.4% |
| randwrite nj=8 | IOPS / mean lat | 8,423 / 947µs | 8,400 / 950µs | −0.3% / +0.3% |
| seqread nj=1 (zero-IPC fast path, warm) | lat | 0.84µs | 0.82–0.84µs | even |

Tail (p99/p99.9) wobbled ±10–20% in both directions across workloads — noise on this box, no directional pattern.

Pre-existing issues found while benchmarking (reproduced identically at the base commit, unrelated to this PR):
1. The POSIX interceptor does not export `statx()`; fio-3.36 sizes files with it, so fio always believes `clio::` files are absent and re-lays-out every read job's files.
2. 8-thread concurrent 4k **sync** writes (`CLIO_CFS_ASYNC_WRITES=0`) fail with EIO ~4.5MB into each file; with async writes (default) the failures latch and surface at the next fsync instead.
3. Any client that forks after `CLIO_INIT` (fio's default process-per-job) hangs on the SHM path — responses are addressed to the child pid's response ring, which was never created. fio needs `--thread`.

PR #852, opened 2026-07-28
Escape the SHM response wait when the server dies mid-flight

### LINKED ISSUE #851: Cluster Tests workflow has been 100% cancelled (30/30 runs): reconnect test client hangs in WaitForLocalServer past its own 30s timeout
## Description

Every one of the last 30 `Cluster Tests` workflow runs (all branches) ended `cancelled` — the job runs ~2.5 h until the workflow timeout kills it. The gate has been effectively dead for days and no longer gates anything.

## Evidence (run 30369661052, dev @ a9f6bc96, 2026-07-28)

The job progresses through the compose suites until the **reconnect** integration test:

```
14:55:47 ClientInit Config at: .../test/integration/reconnect/clio_conf.yaml
14:55:47 IpcManager::ClientInit: retry_timeout = 5s
14:55:47 [INFO] Initializing Clio client for Reconnect tests...
14:55:47 WaitForLocalServer Waiting for runtime via lightbeam (timeout=30s)
17:12:52 ##[error]The operation was canceled.        <- 2h17m later, nothing in between
```

Two independent problems visible:
1. The runtime the reconnect client waits for never comes up (or dies silently) — possibly the stale `restart: true` state issue #725 describes (reconnect is exactly the suite that restarts daemons), or a ServerInit stall in the docker environment. Note this run already contains the #822 inline-retry fix.
2. The client's own `WaitForLocalServer (timeout=30s)` never fires — no timeout error is ever printed, so the 30 s bound is not actually bounding whatever it blocks on (ZMQ recv?), turning a dead daemon into an infinite hang instead of a clean failure.

Fixing (2) alone would convert the 2.5 h workflow cancellation into a fast, diagnosable test failure.

## Acceptance criteria
- [ ] `WaitForLocalServer` actually enforces its 30 s timeout (client exits with the existing error path instead of hanging)
- [ ] Reconnect suite either passes or fails fast with daemon-side logs
- [ ] Cluster Tests workflow completes (pass or fail) instead of being cancelled

lukemartinlogan:
## Summary
- `IpcCpu2Cpu::RecvOut`'s wait loop now re-checks `server_alive_` each pass (as its ZMQ twin always did) and hands the future to `IpcCpu2CpuZmq::RecvOut`, whose `kClientShm`-origin head implements the reconnect/failover + resend path.

## Motivation
Fixes #851 — the root cause of the Cluster Tests workflow being cancelled on 30/30 recent runs (~2.5 h each). A task in flight over SHM when the runtime dies could never complete, and the SHM wait loop had no liveness check and no deadline (default `Wait()`), so the client parked forever. The reconnect integration test hits this deterministically: `AsyncStopRuntime`, then submit — the `IpcManager::Recv` dispatch can win the race against the 1 s heartbeat and choose the SHM path, after which the death is never noticed.

Reproduced and validated against a live 4-node docker cluster (deps-cpu, DooD):
- **Before**: `clio_run_reconnect_tests '[reconnect]'` hangs ≥300 s at "Step 3: Creating pool on new host"; wedged client's main thread parked in `epoll_wait`; no reconnect log lines ever appear.
- **After**: the escape fires ("server died mid-wait; handing off to reconnect path"), `WaitForServerAndReconnect` fails over (`Trying 8/16: 172.26.0.13` → connected), post-failover task completes — suite passes 1/1.

## Test plan
- [x] Live 4-node reconnect suite passes (was: infinite hang)
- [ ] Cluster Tests workflow on this PR completes instead of being cancelled
- [ ] No regression in unit gates

## Breaking changes
None.

### DIFF
--- a/context-runtime/include/clio_runtime/ipc/ipc_cpu2cpu_impl.h
+++ b/context-runtime/include/clio_runtime/ipc/ipc_cpu2cpu_impl.h
@@ -9,6 +9,7 @@
 #define CLIO_RUNTIME_INCLUDE_IPC_CPU2CPU_IMPL_H_
 
 #include "clio_runtime/ipc/ipc_cpu2cpu.h"
+#include "clio_runtime/ipc/ipc_cpu2cpu_zmq.h"
 
 #include <unordered_map>
 
@@ -157,6 +158,19 @@ bool IpcCpu2Cpu::RecvOut(IpcManager *ipc,
       if (start.GetUsecFromStart(now) >= static_cast<double>(max_sec) * 1e6) {
         return false;
       }
+    }
+    // Server-death escape (issue #851): unlike the ZMQ twin of this loop, the
+    // SHM wait used to have NO liveness check — a task in flight over SHM when
+    // the runtime died (reconnect test: AsyncStopRuntime, then submit) left
+    // the client parked here forever, since a dead server can never set
+    // FUTURE_COMPLETE. The 1s heartbeat flips server_alive_; hand the future
+    // to the ZMQ RecvOut, whose kClientShm-origin head implements the
+    // reconnect/failover + resend path.
+    if (!ipc->server_alive_.load(std::memory_order_acquire) &&
+        !ipc->reconnecting_.load()) {
+      HLOG(kWarning,
+           "Recv(SHM): server died mid-wait; handing off to reconnect path");
+      return IpcCpu2CpuZmq::RecvOut(ipc, future, max_sec);
     }
   }
 


lukemartinlogan:
CI-side validation via manual dispatch (Cluster Tests only triggers on push to main/dev, never on PRs): run [30393045231](https://github.com/iowarp/clio-core/actions/runs/30393045231) on this branch **completed in ~18 minutes** — versus 30/30 prior runs cancelled at the 2.5 h workflow timeout. The reconnect suite and all previously-wedged suites pass.

The run then reached later steps that have never executed in CI before (the hang always killed the run first) and surfaced two pre-existing/orthogonal failures — 5/14 CTE distributed SHM-cache tests (incl. the #839 transform-bit tests' first-ever distributed execution) and a cross-node SHM gate exiting 127 (missing binary in the harness). Filed as #853; out of scope here.

Also worth noting for reviewers: because the workflow never runs on PRs, cluster regressions land silently — adding a PR-path trigger (or a light subset) might be worth a follow-up.

Issue #859, opened 2026-07-29
CTE: pinned zero-copy read views of RAM-resident blobs — FAISS IVF ChiMod is copy-bound (resolve exists, view is near, pin is missing)


lukemartinlogan:
From Carmen Vázquez's report *"Improving FAISS IVF Search using CLIO CTE"* (2026-07-22), updated against current `dev` — the report's background section predates several changes that landed this month, so this issue keeps only what is still unsolved.

## Context

A FAISS IVF ChiMod stores each inverted list as its own CTE blob (RAM tier capped, NVMe below), replacing FAISS `OnDiskInvertedLists`' single mmap'd `.ivfdata` file. Searches resolve the needed lists, issue pipelined `AsyncGetBlob()`s, and scan each list as its read completes.

That design already achieved its goal — it eliminates page-cache thrashing in the index ≫ RAM regime:

| Index size | nprobe | CTE QPS (cold/warm) | mmap QPS (cold/warm) | mmap page-ins/pass |
|---|---|---|---|---|
| 4.8 GB | 64 | 148 / 145 | 28 / **238** | 46k / <700 |
| 24 GB | 128 | 21 / 27 | 11 / **93** | 213k / <200 |
| 49 GB | 256 | **8.9 / 7.8** | 2.0 / 5.5 | 757k / ~535k |
| 87 GB | 256 | **4.0 / 3.9** | 0.4 / 0.4 | 2.8M / ~2.7M |

(500 queries, K=10, nprobe=nlist/64, 8 threads, 48 GB node.)

## The remaining bottleneck: one copy per scanned list

Above the crossover CTE wins decisively, but *below* it the warm mmap baseline is faster — and the gap is exactly one memcpy. Every list read, even when the blob is resident in the RAM tier, lands in a temporary buffer before the scanner may touch it. Three observations pin the search as copy-bound, not compute-bound:

- ChiMod timers: waiting on reads > time scanning.
- Switching the scan from scalar to AVX-512 changed throughput by ~nothing.
- Warm mmap (no copy — the page *is* the final location) beats CTE-in-RAM (copy) at equal residency.

## What already exists (do not re-build)

The report listed three blockers; two of them are gone on current dev:

[...]

PR #861, opened 2026-07-29
fix(cte): reader-pin/drain — extent-freeing mutators must wait for in-flight reads (#753 reader half)

lukemartinlogan:
Closes the **reader half of #753** (surfaced while scoping #859's pin requirement).

## The hole

#753's fix serialized the *mutators*: ReorganizeBlob holds the per-blob write token for its whole move, keeps the `BlobInfo` alive (no del+reput window), and GetBlob's torn-layout guard handles mid-mutation snapshots. All four mutators (Put/Reorganize/DelBlob/Truncate) take the token.

But readers deliberately hold **no lock during I/O**: `GetBlobImpl` snapshots `blocks_`, then co_awaits one bdev read per block. An extent-freeing mutator that takes the token *after* the snapshot frees the snapshot's extents mid-`ReadData`; once churn re-allocates them, the read returns **another blob's bytes with rc=0**. Nothing waits for readers, so nothing prevents it. The same hole exists in `FlushData` (which read `blocks_` directly and freed volatile blocks with *no token at all*) and `SemanticSearch` (unpinned, unsnapshotted read). #820's Evict routes through DelBlob, so capacity eviction is one of the triggers.

## The fix — readers pin, extent-freeers drain

A reader-pin word on `BlobInfo` (bit 63 = drain flag, bits 0–62 = pin count; same `atomic_ref` discipline as `write_owner_`):

- **Readers** (`GetBlobImpl` scalar+vectored, `SemanticSearch`, `FlushData`'s read) pin around snapshot+read. `TryPinRead` **never waits while pinned** — it backs off and retries un-pinned — so the protocol is deadlock-free by construction: drainers wait only on pin holders, pin holders wait only on bdev I/O.
- **Extent-freeing mutators** drain pins to zero under the write token before freeing: `ReorganizeBlob` before `ClearBlob`, `DelBlob` before `FreeAllBlobBlocks` (covers #820 Evict), `ResizeBlob`'s shrink (the chokepoint shared by Truncate *and* PutBlob-replace), and `FlushData` before its volatile-free — which now also takes the write token, fixing its pre-existing dangling-`blocks_` read.
- **PutBlob never drains** — it modifies data in place but frees nothing — so read/write concurrency on a blob is unchanged.
- RAII guards (`BlobReadPinGuard`, `BlobReaderDrainGuard`) release on every coroutine exit path; `placement_gen_` remains the independent client-side SHM protection.

This is also the foundation #859 needs: a zero-copy scan is "pin held across compute" instead of "pin held across memcpy".

## Tests (all in `test_reorganize_blob`, suite 9/9 green)

1. **Protocol invariants (#753)** — deterministic white-box: pins block the drain, draining blocks new pins, end-drain re-admits, guards release, plus a 4-reader × 200-drain-cycle cross-thread hammer asserting quiescence inside every drain. **Negative-verified**: stubbing the drain wait fails this test immediately.
2. **Pipelined reads survive extent reuse (#753)** — adversarial integration: 8MB victim seeded as 128×64KB *disk* blocks (multi-block ⇒ each read spans 128 awaits; disk ⇒ forces the RPC path — DRAM reads go via the SHM fast path, which is separately gen-protected), disk tier filled to 100%, reads streamed across a disk→DRAM move, churn burst fired at the exact post-`ClearBlob` mirror publish so it re-allocates the freed extents. Diagnostics confirmed the recipe reaches the window (reads overlap the move; churn lands on the victim's exact freed offsets).
3. **DelBlob under pipelined reads** — rc=0 must mean intact bytes; rc≠0 (blob gone) is legal; garbage never.

Honest caveat: on the 2-worker test fixture the client-driven schedule could not force the final pre-fix straddle (the mover always outpaced in-flight readers), so the deterministic negative lives in the protocol test; the integration tests prove the paths and invariants under real concurrency. The window is wider on multi-worker deployments and via FlushData, which had no protection at all.

Regression battery green: reorganize ×9, tag ops, SHM-cache functional ×5, vectored ×4, evict ×2, query.

### DIFF
--- a/context-transfer-engine/core/include/clio_cte/core/core_tasks.h
+++ b/context-transfer-engine/core/include/clio_cte/core/core_tasks.h
@@ -852,6 +852,22 @@ struct BlobInfo {
   // here — it would deadlock the single worker the instant the holder suspends
   // at a co_await. 0 == unlocked; otherwise a non-zero per-task owner token.
   clio::run::u64 write_owner_;
+  // Reader-pin word (issue #753, reader half). GetBlob deliberately reads with
+  // no lock held: it snapshots blocks_ and then co_awaits bdev reads, so a
+  // concurrent ReorganizeBlob/DelBlob/Truncate that takes the write token
+  // AFTER the snapshot could free the snapshot's extents mid-read -- the
+  // reader would return reused bytes with rc=0. This word closes that hole
+  // without serializing readers against PutBlob (which modifies data in place
+  // but never frees extents): readers hold a PIN for the duration of
+  // snapshot+read, and only the extent-FREEING mutators drain pins to zero
+  // (under the write token) before freeing.
+  //
+  // Layout: bit 63 = drain flag (an extent-freeing mutator is waiting/active);
+  // bits 0..62 = pin count. Readers must NEVER wait while pinned (they back
+  // off and retry un-pinned), which makes the protocol deadlock-free: the
+  // drainer waits only on pin holders, and pin holders only wait on bdev I/O.
+  // Same atomic_ref discipline as write_owner_ above.
+  clio::run::u64 read_state_;
   // Maintained mirror of sum(blocks_[i].size_). GetTotalSize() returns this in
   // O(1) instead of an O(blocks) sum; every blocks_ mutation MUST keep it in
   // sync (ExtendBlob updates it incrementally; cold paths call
@@ -910,6 +926,7 @@ struct BlobInfo {
         trace_key_(0),
         preallocated_size_(0),
         write_owner_(0),
+        read_state_(0),
         total_size_cache_(0),
         placement_gen_(0) {
     prealloc_lock_.Init();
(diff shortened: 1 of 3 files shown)


lukemartinlogan:
CI Windows root-caused and fixed (7e5e3e78): the `cte_reorganize_force_net` variant runs this same binary under CLIO_FORCE_NET, where the new race tests' fixed-count tier fills and multi-round churn hit two environmental corners: (1) tier residue from earlier cases — fills are now adaptive (put until allocation fails); (2) repeating the near-full churn cycle drives ReorganizeBlob into its triple-placement-failure exit, which **permanently loses the blob's data** — a real capacity-pressure bug, now filed as #863 with the repro from these tests. The race runs one fully-armed round with churn capped at half the freed capacity to stay out of that corner until #863 is fixed. Also fixed a scheduling flake in the protocol hammer (bounded wait for reader progress per drain cycle).

Suite is 9/9 green locally in BOTH normal and CLIO_FORCE_NET=1 modes.

Issue #867, opened 2026-07-29
Cluster Tests workflow has not been green since Jul 18: cross-node SHM gate exits 127, 4-node SHM-cache puts fail, and auto-cancel masks it all


lukemartinlogan:
## Summary

The **Cluster Tests** workflow has zero successful runs since **2026-07-18**. Nearly every run is auto-cancelled (dev pushes land faster than the ~20-min suite; concurrency kills the in-flight run), and the only two runs that COMPLETED in that window — run 30393045231 (Jul 28, workflow_dispatch) and run 30421640132 (Jul 29, dev push / PR #779) — both failed with the **identical, deterministic** signature. This currently blocks the dev→main promotion (PR #779, where it is the sole remaining red check after icx flake reruns passed).

## Failure 1: cross-node SHM client gate — node1=node2=exit 127 within ~6s

The "tiny cross-node SHM gate" (grayscott_congestion, added Jul 19 in 5444ef78/0456d7d7 — i.e. the workflow has NEVER been green with this gate in it) fails with both containers exiting **127** right after their `sleep 6`, before any real work:

```
phase 'tiny cross-node SHM gate': node1 exit=127 node2 exit=127
✗ Cross-node SHM client gate FAILED (node1=127 node2=127)
```

127 = command-not-found / loader failure for `/workspace/build/bin/clio_run` or `test_grayscott_congestion` inside the containers. The binary IS built in the shared build step (job log: `Built target test_grayscott_congestion`), and the sibling distributed-phase containers use the same `${HOST_WORKSPACE:-${IOWARP_CORE_ROOT:-/workspace}}` mount successfully in the same job — so the specific cause is not yet provable from CI logs, because the gate's log filter (`grep -aE "gs-congest|OUTPUT|RESULT|STALL"`) matches NOTHING the containers actually print (their lines are prefixed `gs-nodeN:` / `+`). **PR with the observability fix: the next failing run will show the raw error.**

Candidate causes to check once logs are visible: mount env resolution in the gate step (it exports only HOST_UID/GID; run_tests.sh falls back to computed REPO_ROOT), loader path (`LD_LIBRARY_PATH` override in the gs compose), or a compose YAML folding artifact in the chained command.

## Failure 2: 4-node distributed — two SHM-cache subtests fail on Put

In the "CTE distributed test (4-node, shm/tcp/ipc)" step, the same two subtests fail on both completed runs:

```
[...]

PR #868, opened 2026-07-29
ci(cluster): dump raw container logs when the gs congestion gate fails (#867)

lukemartinlogan:
## Summary
- **Fixes the exit-127 root cause**: `grayscott_congestion/docker-compose.override.yml` — explicitly headed "LOCAL-ONLY override (untracked)" — was accidentally committed (430040ac, Jul 21). Compose auto-loads it, and it bind-mounts a one-dev-box Docker-Desktop path (`/host_mnt/c/.../build-boost`) over `/workspace/build`, which is empty on CI runners → `clio_run` / `test_grayscott_congestion` not found → both nodes exit 127 on every completed Cluster Tests run since. Deleted, and `docker-compose.override.yml` gitignored so per-box overrides stay local.
- The cross-node SHM gate's log filter greps for `gs-congest|OUTPUT|RESULT|STALL`, but the containers print lines prefixed `gs-nodeN:` and `+` (set -x) — no actual output ever matched, so every CI failure showed two empty sections and bare exit codes.
- Match the real `gs-node` prefix, and on any non-zero node exit dump the raw 40-line log tails of both containers.

## Motivation
Part of #867: the gate has failed `node1=node2=127` on every completed Cluster Tests run since it was added (Jul 19), and the cause is unprovable from CI logs because the error lines are filtered out. This makes the next failing run self-diagnosing.

## Test plan
- [x] `bash -n` clean; failure path exercised locally (container logs dumped on exit 127)
- [ ] Next Cluster Tests run shows the raw error for #867


lukemartinlogan:
Cluster Tests triage for run 30468105512 (latest cycle on this branch), from the CI-watch loop:

- **Every sub-suite passed**: ManyToOne (bdev batch + cfs deferred append), AllToOne barrier, Distributed (shm/tcp/ipc), Induction, Migration (re-routing + broadcast), and all 8 FUSE integration tests.
- **The only failure is the leader-election step**: `##[error] The action 'Runtime leader-election test (leader death + re-election)' has timed out after 15 minutes.` Timeline: leader dies → client retries the dead segment for 5 s → host sweep → **successfully reconnects to the new leader** `172.24.0.13` (generation=835406895771) at 16:07:43 — then total silence until the 15-min step timeout at 16:22:28. The client is wedged *after* a successful reconnect, waiting on something the new leader never completes (in-flight task re-issue? a future bound to the dead server's generation?).
- The `RouteTask: RouteLocal returned 4 ... method=14/21` burst at the log tail is unrelated teardown noise from the FUSE test's shutdown (31 lines, after 'All FUSE integration tests passed' and `DestroyLocalPool (1,0)`): the periodic admin pumps bouncing off their just-destroyed pool during cleanup.

lukemartinlogan:
Dev Cluster Tests triage (run 30477668601, post-#861 merge — but the failure predates it: dev's 04:12 run failed identically before #861 existed on dev):

The leader-elect suite gets further than the 15-min hang triaged earlier — Phase 1 (leader kill) and the restart complete, and the post-restart health battery actually runs. It then fails 5 tests, all on **`PutBlob rc != 0`**, and the cause is one line earlier in the daemon output:

```
core_runtime.cc:656 WARNING Create: No storage devices configured
```

The **re-created CTE pool after the leader restart comes up with ZERO storage targets**, so every put in the battery fails (the failing tests are just whichever ones write first). Either the restart path re-creates the CTE container without re-parsing its storage config, or the leader-elect harness's restart invocation doesn't propagate the compose config (`clio_conf.yaml`) to the restarted daemon. The 'Cross-node SHM client gate FAILED (node1=127 node2=127)' later in the job is downstream of the same broken post-restart state.

PR #870, opened 2026-07-29
test(cte): scale #753 race geometry under CLIO_FORCE_NET — fixes dev's icx timeout

lukemartinlogan:
dev's post-#861 CI Windows failed on `cte_reorganize_force_net` (Timeout) on the slowest icx config: the #753 race tests' seeding runs every op through the ZMQ loopback in the force_net variant, and Windows loopback latency blew the 300s default (the PR-branch runs passed with less headroom to spare).

Fix per the `cte_block_reuse_force_net` precedent: the geometry shrinks under `CLIO_FORCE_NET` (2MB/32-chunk victim, halved read pipeline — the reader-pin logic under test is transport-independent), plus a 600s timeout for the variant. Local: 9/9 both modes, force_net suite ~6s wall.

Note: dev's Cluster Tests red is a separate, pre-existing issue (post-restart CTE pool comes up with no storage devices — triaged on #868).

### DIFF
--- a/context-transfer-engine/test/unit/CMakeLists.txt
+++ b/context-transfer-engine/test/unit/CMakeLists.txt
@@ -861,7 +861,11 @@ add_test(NAME cte_reorganize_all
 set_tests_properties(
     cte_reorganize_all
     PROPERTIES
-        TIMEOUT 300  # 5 minute timeout
+        # 600s: the #753 race tests seed a multi-chunk victim and fill the
+        # disk tier; Windows Debug runners need the headroom even with the
+        # in-test geometry scaling (cte_reorganize_all timed out at 300s on
+        # windows-2025).
+        TIMEOUT 600
         LABELS "unit;reorganize;cte"
         ENVIRONMENT "LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/bin:$ENV{LD_LIBRARY_PATH};CLIO_TEST_DATA_DIR=${CMAKE_BINARY_DIR};CLIO_BIND_ADDR=127.0.0.1"
 )
--- a/context-transfer-engine/test/unit/CMakeLists.txt
+++ b/context-transfer-engine/test/unit/CMakeLists.txt
@@ -926,7 +926,11 @@ clio_add_force_net_test(NAME cte_tag_force_net
     COMMAND test_tag_operations ENVIRONMENT "${_cte_fn_env_data}")
 clio_add_force_net_test(NAME cte_query_force_net
     COMMAND test_query ENVIRONMENT "${_cte_fn_env_data}")
-clio_add_force_net_test(NAME cte_reorganize_force_net
+# 600s: the #753 race tests seed a multi-chunk victim + fill the disk tier,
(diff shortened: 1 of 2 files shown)


lukemartinlogan:
Second fix pushed: the leak-check job's `Subprocess aborted` was the protocol hammer calling `std::terminate` — a REQUIRE tripping while 4 joinable reader threads were alive (per-cycle progress can stall on the LSAN-slowed 2-vCPU runner). Assertions now run only after stop+join, with aggregate (not per-cycle) progress requirements. 5× force_net stress + both full-suite modes green locally.

lukemartinlogan:
Data point from the first dev run containing #861 (merge a4525115, icx Windows): the reorganize suite now TIMES OUT in both icx jobs — `cte_reorganize_force_net` on windows-2022 and plain `cte_reorganize_all` (plus `cr_compose_force_net`) on windows-2025. The 600s force_net timeout bump here covers the force_net variant, but `cte_reorganize_all`'s timeout on the plain path suggests the #861 pin/drain additions slowed the suite generally on Windows — worth extending this PR's timeout/geometry treatment to the non-force_net variant too (or trimming the new race tests' Windows cost). Rerunning the failed dev jobs meanwhile.

lukemartinlogan:
Follow-up: after a rerun, the reorganize suite has now timed out in **3 of 3 icx dev job executions** (plain `cte_reorganize_all` twice on windows-2022/2025, `cte_reorganize_force_net` once) — deterministic post-#861 Windows slowness, not the rotating flake. Reruns won't clear it: this PR should raise/scale the PLAIN `cte_reorganize_all` timeout alongside the 600s force_net bump (or gate the new race-test geometry down on Windows). Until then every dev push will fail both icx jobs.

lukemartinlogan:
Scope grew: dev (which has #861 but not this PR) now shows **three** reds this PR fixes — the original icx force_net timeout, the leak-check protocol-hammer abort, and (new tonight) **windows-2025 timing out the plain `cte_reorganize_all` at 300s** — Windows Debug is slow enough without any loopback involved. Latest push scales the race geometry on `_WIN32` builds as well as under CLIO_FORCE_NET and gives `cte_reorganize_all` the same 600s headroom as its force_net variant. Both full-suite modes 9/9 locally. **Merging this PR heals all three dev gates.**

lukemartinlogan:
Latest push status: leak-check ✅, icx windows-2022/2025 ✅ (timeouts cured) — but **icx (windows-2025-vs2026)** now shows a real failure instead of a timeout: `ReorganizeBlob - Reader pin/drain protocol invariants (#753)` fails `REQUIRE(quiescence_violations == 0)` at test_reorganize_blob.cc:886 (test ran 2.2s, so not load-related). Either the invariant checker still has a window ed8b36b6 didn't close on that toolchain's timing, or vs2026 is exposing a genuine pin/drain quiescence hole. Job: https://github.com/iowarp/clio-core/actions/runs/30513293131 — full subtest output is in the icx (windows-2025-vs2026) log around 04:10:09.

PR #869, opened 2026-07-29
ycsb: async write window — workload A to Redis parity+ (2–3.7x), F ahead of Redis (1.9x)

lukemartinlogan:
Stacked on #864. Answers the 'why is A slow' question: every update was one **blocking** PutBlob round trip (the same depth-1 cliff `clio_cte_bench` shows: 17.8k→51k ops/s from depth 1→128), while Redis updates a single hash field server-side.

Writes now pipeline through the private-memory `AsyncPutBlob` (#830 — client mode stages the bytes at submit, so the JNI array releases immediately and the task frees its own staging) into a shared in-flight window: `CLIO_YCSB_WINDOW` (default 64, `0` = old synchronous behavior), oldest-future reap outside the lock, sticky errors, `cleanup()` drains so no put escapes its phase. Reads may observe the pre-put value of an in-flight key — YCSB core workloads don't verify read values, and the phase boundary drains.

**Measured** (CLIO_PREFAULT=1GB, same harness/box, run-phase ops/s):

| Workload | sync (prefault) | async | redis |
|---|---:|---:|---:|
| A (50/50) | 9,820 | **21,245 / 36,049** (two runs) | 21,650 |
| F (RMW) | 10,708 | **20,202** | 14,762 |
| C (read-only) | 86,059 | 46,729–59,559 | 21,044 |

C's spread is this box's known 46–86k run-to-run variance (read path untouched); loads were inconclusive under the same noise. Multi-run averaging on quiet hardware before publishing.

Remaining update-side lever (not in this PR): eliminating the client-side read-modify-write via partial `PutBlob`-at-offset or a server-side field merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Final results (2026-07-31, AsyncMultiPutVectored) — batching breakthrough

Supersedes the tables below. The deferred pipeline now ships 64-put batches
(Method::kMultiPutBlob, commit a8cbb274): AsyncPutBlobDefer accumulates puts
into a 256KB SHM chunk — read-your-writes serves from the ACCUMULATING batch
pre-flush — and one task executes the batch as nested PutBlob calls. Still
the unbounded-until-SHM default; every run verified full-load / zero-error /
zero-NOT_FOUND.

**2M ops (E: 100k), k ops/s — both clio modes verified full-load /
zero-error / zero-NOT_FOUND:**

| Workload | clio embedded | clio external (client<->daemon) | RocksDB (embedded) | Redis (server) |
|---|---:|---:|---:|---:|
[...]

PR #878, opened 2026-07-31
cte: deferred-put pipeline with 64-put batching — clio leads RocksDB on A/B/C/D in both deployment modes

lukemartinlogan:
Twelve commits building the deferred-put pipeline end to end (818dd3c0..a8cbb274), stacked on the merged #869.

## Result (YCSB, 2M ops, 100k x 1,124B records, 4 threads; every clio run verified full-load / zero-error / zero-NOT_FOUND)

| Workload | clio embedded | clio external (client<->daemon) | RocksDB (embedded) | Redis (server) |
|---|---:|---:|---:|---:|
| A (50% update) | **118.6k** | 116.2k | 117.1k | 22.2k |
| B (95% read) | **794.9k** | 418.5k | 255.3k | 21.7k |
| C (100% read) | **1,298.7k** | 856.5k | 312.7k | 21.1k |
| D (read latest) | **894.1k** | 480.3k | 434.3k | 22.9k |
| E (scan) | 14.7k | 14.2k | **18.5k** | 0.44k |
| F (RMW) | 104.7k | **104.7k** | 136.1k | 14.7k |

Embedded clio leads RocksDB on A/B/C/D (1.01x / 3.1x / 4.2x / 2.1x). The external column is the structural headline: 64-put batches collapse 64 IPC crossings into one, so the separated deployment reaches **embedded parity on A and F** and beats RocksDB on B/C/D — from across a process boundary. All under unbounded-until-SHM async semantics with strict read-your-writes (workload D's read-latest pattern re-verified at zero NOT_FOUND; workload C previously byte-verified under YCSB dataintegrity).

## What's in it

- **bdev**: 512B/1KB/2KB block classes + 512B RAM-bdev allocation granularity (a ~1.1KB blob cost a full 4KB block; a 256MB tier silently 'filled' at 65,536 blobs with silent put failures — root cause of the earlier invalidated results).
- **Deferred-put pipeline as core client APIs** (no binding policy, no env knobs): `AsyncPutBlobDefer` owns an SHM copy at submit (caller buffers free immediately, both modes), grows until shared memory is exhausted then awaits oldest puts; optional `max_inflight_bytes` pacing as a METHOD parameter. `AsyncGetBlobDefer` serves read-your-writes by composing from the newest-wins union of ALL pending extents per key (sharded, allocation-free FNV-keyed tables) — a put task carries its own bytes, so serving is mode-independent.
- **`AsyncMultiPutVectored` (Method::kMultiPutBlob)**: the pipeline's batch output — puts accumulate into a 256KB SHM chunk (extents serve reads pre-flush) and ship as ONE task executing up to 64 nested PutBlob calls. Amortizes per-put task machinery AND dissolves deep-backlog congestion; this is what removed the depth-vs-throughput tradeoff (A went 22k -> 118.6k at full scale).
- **`CLIO_RUN_INLINE`**: general inline-execution macro (runtime + local + worker fiber + not-unit-test; `simple_test.h` stamps `CLIO_UNIT_TEST`); adopted by bdev write/read/alloc/free. Measured throughput-neutral at both depths — kept for latency paths and generality, numbers in the commit.
- **Shard-ingest drain budget 16 -> 256** (+12% client-mode A, measured back-to-back).
- Correctness fixes found by the workload-D magnifying glass along the way: mid-reap pending-key race, stale zero-size mirror records in `Tag::GetBlobSize`/`TryGetBlobViewShm`, priv::string allocator construction in task ctors, and a flush-under-registry-lock deadlock.
- YCSB binding reduced to a plain caller of the core APIs; Python bindings expose the defer APIs.

Benchmark methodology, correction history, and per-build tables are archived in the merged #869's body/comments.


lukemartinlogan:
CI-loop note: **GPU Tests (CUDA) is failing on this branch with a real nvcc compile error**, not a flake — run 30611348734, `Build CUDA conda package (compile only)`:

```
run_inline.h(86): error: function "clio::run::Container::Run" returns incomplete type "clio::run::TaskResume"
run_inline.h(80): error: class "clio::run::bdev::AllocateBlocksTask" has no member "BeginRunContext"
run_inline.h(83): error: ... no member "RunFuture"   (same for FreeBlocksTask, WriteTask, ...)
ipc_manager.h(2053): error: ... no member "RunCtxPtr"
```

Looks like the new `CLIO_RUN_INLINE` header (7c2a4245) relies on task members/types that aren't visible in CUDA translation units — either gated behind a host-only macro (remember: use `CTP_IS_GPU`/`CTP_IS_HOST`, not raw `__CUDACC__`) or an include-order/forward-decl issue (`TaskResume` is incomplete at the point `Container::Run` is instantiated).

Also FYI: this branch predates the #875 merge, so once you rebase/merge dev you'll pick up the pinned py312 Miniconda for the conda legs (the runner image's conda base moved to Python 3.14 on 07-30 and crashes conda-build).

lukemartinlogan:
CI-loop follow-up: **CI Adapters is also failing on this branch with a real functional regression** (linux all-5 and macos hdf5_vol legs, run at 07:10Z):

```
clio_cte_posix_unit_tests ***Failed
core_runtime.cc:5321 ERROR ModifyExistingData: WRITE FAILED - task[0] wrote 0 bytes, expected 1048576
(repeats for 1024 / 4 / 4103-byte writes across test_posix_adapter.cc:164/233/300/...)
```

Every modify-existing write path is returning 0 bytes written — distinct from the nvcc/run_inline.h compile break above, and it smells like the new write-path work on this branch (AsyncMultiPutVectored batching / drain-budget / inline-run changes) breaking the bdev write task's byte accounting or its segments path. Both failures are deterministic-looking, not runner flakes.

lukemartinlogan:
The GPU Tests failure is unchanged across the last three pushes, so here is the concrete fix. `task.h` declares `BeginRunContext` / `RunFuture` / the coroutine machinery (and `TaskResume` as a complete type) only under `#if CTP_IS_HOST` (task.h:306-337, 1266+), but `run_inline.h`'s `RunInlineOrSend` uses them unguarded — so every nvcc TU that includes it fails. Gate the inline path the same way task.h gates the members:

```cpp
template <typename TaskT>
inline clio::run::Future<TaskT> RunInlineOrSend(
    clio::run::shared_ptr<TaskT> &task) {
#if CTP_IS_HOST
  ... existing body ...
#else
  // Host-only fiber machinery (BeginRunContext/RunFuture/TaskResume) is
  // excluded from GPU-compiler passes; inline execution is meaningless there.
  return CLIO_CPU_IPC->Send(task);
#endif
}
```

(or guard the whole header body after the includes with `#if CTP_IS_HOST` and provide the `CLIO_RUN_INLINE` macro as plain `Send` in the `#else`).

Issue #818, opened 2026-07-23
CTE SHM cache: compressed blobs are indistinguishable from raw ones in ShmBlobRecord


lukemartinlogan:
## Summary

The #783 shared-memory cache lets a client copy a blob's payload straight out of the RAM
bdev segment. `ShmBlobRecord` carries **no compression state**, so a compressed blob is
indistinguishable from a raw one: the record's `total_size_` is the *stored* size, the bytes
behind it are `CompressionHeader + codec output`, and byte offsets into them are meaningless.
Today that happens to be consistent with the RPC path (core `GetBlob` also returns stored
bytes), so this is **not a live data-corruption bug** — it is a latent one, and a missed
opportunity.

## What the runtime knows and the cache drops

`ContentTransferEngine`'s compressor chimod stores a compressed blob as an ordinary CTE blob
(`context-transfer-engine/compressor/src/compressor_runtime.cc:773`):

```
PutBlob(tag, name, offset, sizeof(CompressionHeader) + compressed_size, ...)
```

with a `CompressionHeader{lib, preset, original_size}` prefix. The core runtime records the
provenance on the blob (`core_runtime.cc:1624`, under `#if CTP_ENABLE_COMPRESS`):

```cpp
blob_info_ptr->compress_lib_    = context.compress_lib_;
blob_info_ptr->compress_preset_ = context.compress_preset_;
```

But `Runtime::BuildShmBlobRecord` (`core_runtime.cc:~300-360`) copies only sizes, timestamps,
score and the block descriptors. `kShmBlobDirectReadable` is set purely on "every block is
node-local and `kRam`" — compression is not consulted. `ShmBlobRecord`
(`core/include/clio_cte/core/shm_metadata_cache.h`) has no field for it.

Consequences:

1. **`Tag::GetBlobSize` fast path reports the compressed size.** Consistent with RPC today,
   but it is not the number a caller of a compressed blob wants, and nothing marks it.
2. **Offsets are silently wrong for any partial read.** `TryReadBlobShm(tag, name, out, size,
   offset)` indexes into codec output. It returns `true`. Again the RPC path is equally
   wrong today — which is precisely why the hazard is invisible.
3. **The moment core `GetBlob` learns to decompress transparently, the SHM path silently
   diverges** and starts handing back compressed bytes with a success return. That is a
   correctness landmine sitting under a "pure optimisation" fast path whose whole safety
   argument is "it can only be faster or equivalent, never wrong".
[...]

Issue #856, opened 2026-07-29
Leader recovery crashes the new SWIM leader with free(): invalid pointer during container redistribution — wedges Cluster Tests leader-election


lukemartinlogan:
## Description

During the leader-elect integration suite, after the old leader's runtime is killed, the newly-elected SWIM leader **crashes with `free(): invalid pointer`** while redistributing the dead node's containers, taking down a second node and wedging the suite (and the Cluster Tests workflow) indefinitely.

## Evidence (local 4-node DooD repro, dev @ 47bdced8, 2026-07-29)

node2 (new leader) log tail, verbatim sequence:

```
SWIM: Node 0 confirmed dead after suspicion timeout
IpcManager: Node 0 (172.24.0.10) marked as DEAD
Recovery: Leader initiating for dead node 0
Recovery: 2 containers to redistribute from node 0
[SendIn] Task ... target node 0 is dead, net_timeout=0 -> skip
[ClientPool] Creating new persistent connection to 172.24.0.11:9413
[ClientPool] Connection established to 172.24.0.11:9413
PoolManager: Updated mapping for pool PoolId(major:70000, minor:0) container 0 -> node 3
PoolManager: Updated mapping for pool PoolId(major:1, minor:0) container 0 -> node 2
free(): invalid pointer
```

The crash lands inside `Runtime::RecoverContainers` (`admin_runtime.cc`) after the second `UpdateContainerNodeMapping` — the next operations for the self-targeted admin-pool assignment are `WriteAddressTableWAL` and the replacement `DynamicContainer` creation, so the invalid free is in that stretch. The runtime is the container's PID 1, so the crash kills the node2 container outright.

## Downstream effects

- Surviving nodes (3, 4) spam `ScanSendMapTimeouts ... timed out waiting for dead node` for tasks addressed to nodes 1 and 2.
- The test client, which had correctly failed over (`ReconnectToNewHost: Connected to 172.24.0.12`), resends its pool-create; the task's admin container now maps to (dead) node 2, no response ever arrives, and the client parks forever — some runtime-side tasks get the #628 network-timeout RC, but the client's future never completes.
- This is what dev run 30409150924's Cluster Tests has been stuck on for 1h+ (`Runtime leader-election test` step; the same step took 51 s on the pre-merge dispatch run 30393045231 — the crash is likely timing-dependent/nondeterministic, as heap corruption tends to be).

## Repro

[...]

PR #880, opened 2026-07-31
test(cte): skip node-local SHM-cache tests on multi-node deployments (#879)

### LINKED ISSUE #879: Cluster Tests (Distributed) red since #818/#783 landed: SHM-cache tests create Dynamic pools with node-local RAM targets
Every completed dev Cluster Tests run since 2026-07-29 fails the Distributed phase with:

- `CTE SHM cache refuses direct reads of transformed blobs` — REQUIRE `put->return_code_ == 0` (test_core_functionality.cc:2978)
- `CTE SHM cache write-then-read cycle benchmark` — REQUIRE `do_write() == 0` (:3128)

Runs: 30421640132, 30477668601, 30508999714, 30516364138, 30576039683, 30620487450 — identical signature; the suite has never been green since the tests landed (8e36b5fb #818, a7638b2a #783). Pre-dates and is unrelated to #878/#876.

**Root cause:** these tests create their isolated CTE pool with `PoolQuery::Dynamic()` (containers on all 4 nodes) but register their only target — a RAM bdev — with `PoolQuery::Local()` (this node only). On multi-node, a put whose blob name hashes to a remote container has no reachable target → PutBlob rc != 0. The sibling payload test (pool 7801/671) has the same structure and passes only because its blob names happen to hash to the local container.

The SHM-cache direct-read fast path is node-local by definition, so multi-node coverage adds nothing here. Fix: skip these tests when `GetNumHosts() > 1` (payload test guarded too — it is one name-hash away from the same failure).

Side note also seen in the same job wave: the two tests share hard-coded PoolId(7802,0)/(672,0) with different pool names — harmless today (second creation returns the existing pool) but worth unsharing if the tests ever diverge.

lukemartinlogan:
Fixes #879 — the standing red on dev's Cluster Tests Distributed phase.

## What was failing
Every completed dev Cluster Tests run since 2026-07-29 failed the same two tests in the Distributed (4-node) phase with `PutBlob rc != 0`:
- `CTE SHM cache refuses direct reads of transformed blobs` (#818)
- `CTE SHM cache write-then-read cycle benchmark` (#783)

## Root cause
These tests build an isolated CTE pool with `PoolQuery::Dynamic()` — a container on **every** node — but register their only target (a RAM bdev) with `PoolQuery::Local()` — **this node only**. On multi-node, a put whose blob name hashes to a remote container has no reachable target and fails. The sibling payload test has the identical structure and passed only because its blob names happen to hash to the local container.

## Fix
The path under test reads payloads out of node-local shared memory, so multi-node runs add no coverage. All three tests (payload included — it is one name-hash away from the same failure) now skip with a logged `[#879]` warning when `GetNumHosts() > 1`; single-node suites keep the full assertions unchanged.

Compile-checked; single-node behavior untouched (guard returns false at 1 host).

### DIFF
--- a/context-transfer-engine/test/unit/test_core_functionality.cc
+++ b/context-transfer-engine/test/unit/test_core_functionality.cc
@@ -2744,7 +2744,16 @@ TEST_CASE("CTE SHM cache metadata read benchmark",
  * there is name-hash luck, not coverage. Skip them; single-node suites keep
  * the full assertions. */
 static bool SkipShmCacheTestOnMultiNode(const char *test) {
-  if (CLIO_IPC->GetNumHosts() <= 1) return false;
+  size_t hosts = CLIO_IPC->GetNumHosts();
+  if (hosts <= 1) {
+    // The external test client never loads the cluster hostfile (its
+    // GetNumHosts() is 1 even on the 4-node suite), so fall back to the
+    // harness's own signal: the distributed docker-compose exports
+    // CTE_NUM_NODES, which this file already consults elsewhere.
+    const char *n = std::getenv("CTE_NUM_NODES");
+    if (n != nullptr) hosts = static_cast<size_t>(std::atoi(n));
+  }
+  if (hosts <= 1) return false;
   HLOG(kWarning,
        "[#879] {}: node-local SHM-cache test SKIPPED on multi-node deployment",
        test);
--- a/context-transfer-engine/test/unit/test_core_functionality.cc
+++ b/context-transfer-engine/test/unit/test_core_functionality.cc
@@ -2735,8 +2735,25 @@ TEST_CASE("CTE SHM cache metadata read benchmark",
  * authoritative RPC read. A fast path that returns the wrong bytes is far
  * worse than no fast path, so correctness is asserted before speed.
  */
+
+/** The SHM-cache direct-read fast path is node-local by definition, and the
+ * isolated pools the payload/transform/W-T-R tests build pair a Dynamic
+ * (every-node) CTE pool with a RAM target registered ONLY on this node. On
+ * multi-node deployments a put whose blob name hashes to a remote container
+ * has no reachable target and fails (#879) -- whether a given test survives
+ * there is name-hash luck, not coverage. Skip them; single-node suites keep
+ * the full assertions. */
+static bool SkipShmCacheTestOnMultiNode(const char *test) {
+  if (CLIO_IPC->GetNumHosts() <= 1) return false;
+  HLOG(kWarning,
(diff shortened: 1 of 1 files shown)


lukemartinlogan:
**Validation complete** — since Cluster Tests only triggers on dev pushes, I dispatched it on this branch three times; each run refined the fix:

1. Run 30624389124: guard never fired — the external test client doesn't load the cluster hostfile, so `GetNumHosts()` returns 1 even on the 4-node suite. → fixed by falling back to the harness's `CTE_NUM_NODES` export (d9b14d6d).
2. Run 30626859523: guard fired (transform/W-T-R now skip) but surfaced the same class in the **metadata read benchmark** — it demands 2000/2000 local SHM hits against the shared Dynamic pool, and a blob whose container lands on a remote node keeps its record in that node's SHM → local miss by placement luck. → guarded it plus the functional test which makes the same must-hit assertions (bd4d613f).
3. Run 30629231673: **zero test failures, all five SHM-cache tests skip cleanly** (5 SKIPPED lines).

The only remaining red in the Distributed job is the *Runtime leader-election* step's 15-minute timeout — pre-existing, unrelated to this diff (test-file-only branch), and now documented with three run links on #856 (it's that issue's post-re-election recovery hang surfacing in CI).

lukemartinlogan:
Direction change per Luke: instead of skipping the node-local SHM-cache tests on multi-node deployments, commit 9fdd53b8 pins every tag/blob task in the five shm_cache tests to `PoolQuery::Local()`. The cache is node-local by design but must keep working on multi-node — it serves LOCAL data wherever it runs, and a remote blob is a clean RPC-served miss. With local routing the puts land on the executing node's target and publish into that node's SHM, so the assertions hold on every node of the 4-node suite (no placement luck, no lost coverage). All five tests pass single-node; dispatched a Cluster Tests run on this branch for the 4-node validation.

lukemartinlogan:
4-node validation: the dispatched Cluster Tests run on this branch is **green** — https://github.com/iowarp/clio-core/actions/runs/30645588266. The previously-failing SHM-cache subtests now pass on the distributed suite with real assertions (local-pinned placement), not skips. With the leader-election step already fixed, this closes out the last #867 Cluster Tests failure once merged.

lukemartinlogan:
**Pin-to-Local validated**: dispatched Cluster Tests on 9fdd53b8 → https://github.com/iowarp/clio-core/actions/runs/30648163036 — **fully green**, the first green Cluster Tests run in this whole arc. All five SHM-cache tests now run (not skip) on the 4-node suite with their assertions holding, and the leader-election step passed this time as well (its earlier 3/3 dispatch timeouts remain tracked on #856 as load-sensitive). This PR is now validated end-to-end on both the unit matrix and the distributed suite.

PR #885, opened 2026-07-31
fix(hdf5): correctness pass on the CLIO VFD and VOL connectors

mattjala:
Both HDF5 connectors were audited against explicit criteria and the problems found were fixed. Most of these were cases where the connector returned wrong data but reported success, preventing anything downstream from noticing a problem. The VFD compared files by name instead of by identity, and treated a short read as the end of the file. The VOL sized its cache from the caller's memory type rather than the type
actually stored on disk, and kept serving a cached copy of a file that had since been truncated. The VOL also reported success for native reads and writes that had actually failed in some of its data paths. Separately, simply having the VOL library sitting in the plugin directory could crash a plain H5Fopen of a missing or non-HDF5 file, because HDF5 asks every plugin it finds whether it can open the file and the connector dereferenced an object that is null on that path. The connector did not have to be selected or requested for this to happen.

Neither connector could previously run without CLIO present; both now fall back to plain pass-through behaviour. The VFD also gains a CLIO_VFD_CACHE environment variable to switch its cache tier off, matching the VOL's existing CLIO_VOL_CACHE, so the cache can be disabled without editing application source -- which matters because the deployment this driver targets sets HDF5_DRIVER on an unmodified program. Test coverage roughly doubled: the VFD suite went from 15 to 21 sections plus a new no-runtime binary, and the VOL compatibility suite from 20 to 23 cases, with three of the fixes verified to fail on the previous code. The READMEs record what the error-path tests exposed, including one gap that is still open: the VOL cache can hide errors in the file itself, since it has no way to know that the file changed outside HDF5.

### DIFF
--- a/.github/workflows/ci-vfd.yml
+++ b/.github/workflows/ci-vfd.yml
@@ -89,10 +89,15 @@ jobs:
                 -DCLIO_CORE_ENABLE_GRAY_SCOTT=OFF \
                 -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
                 -DCLIO_CTP_LOG_LEVEL=1
-              # Targeted build: just the VFD test target pulls in the driver +
-              # the runtime libs it depends on (add_dependencies), not the whole
+              # Targeted build: the VFD test targets pull in the driver + the
+              # runtime libs they depend on (add_dependencies), not the whole
               # adapter/test matrix.
-              cmake --build build-vfd --target clio_cte_vfd_unit_tests -j"$(nproc)"
+              #
+              # Both test binaries must be named here. ctest below selects by
+              # name (-R vfd), so any target left out is still selected and then
+              # reported as "Not Run" 
+              cmake --build build-vfd -j"$(nproc)" \
+                --target clio_cte_vfd_unit_tests clio_cte_vfd_no_runtime_tests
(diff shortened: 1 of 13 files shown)


lukemartinlogan:
Drive-by from CI triage: the CI VFD failure is `clio_cte_vfd_no_runtime_tests (Not Run)` — ctest "Not Run" means the test executable was never built (missing binary / broken target wiring), not an assertion failure. Likely the connector renames/moves in this PR left the `add_test`/target registration for that suite pointing at something that no longer builds. The rest of the matrix is still running.

lukemartinlogan:
The `vfd (linux, native write-through)` failure is not your test logic — `clio_cte_vfd_no_runtime_tests` shows **Not Run** because the binary is never built: `ci-vfd.yml` does a targeted build,

```yaml
cmake --build build-vfd --target clio_cte_vfd_unit_tests -j"$(nproc)"
```

so the new executable your CMakeLists adds isn't produced, and ctest can't find it. Fix is one line in `.github/workflows/ci-vfd.yml` (~line 95):

```yaml
cmake --build build-vfd --target clio_cte_vfd_unit_tests clio_cte_vfd_no_runtime_tests -j"$(nproc)"
```

(The rest of the run was green: `clio_cte_vfd_unit_tests` passed with your changes.)

PR #890, opened 2026-08-03
test(hdf5): merge and expand VOL/VFD test suite

mattjala:
This PR expands the VOL's test suite into a harness that's shared between the CLIO VFD and the CLIO VOL, using the same set of test files. Each case is written and read both ways and compared against what HDF5 produces, and the resultant files are checked with HDF5 CLI tools to verify integrity to ordinary (non-CLIO) readers.

Several kinds of checks were missing and are added here. Byte-identical outputs for connective vs. native written files, error code checking, and direct POSIX integrity checks (to guard against cases where e.g. CLIO intercepts and covers up a problem in the underlying file). Tests for a variety of more advanced HDF5 features were added, including  object and region references, variable-length and opaque and committed datatypes, null dataspaces, external links, object copy and delete, creation-order tracking, more selection shapes, and more filters. Although support for most of these features isn't an immediate priority (and support for some may be out-of-scope entirely), it's good to have a consistent metric by which to measure and document the current leve of support.

The tests also include the ability to report KNOWN-GAPs rather than failures, which is used in one case right now - a test (corrupt_checksum_read) where dataset is written with a fletcher32 checksum, read once through the VOL so the cache holds it, then its bytes are damaged on disk behind HDF5's back. Native HDF5 reads it and correctly fails with a checksum error. The VOL reads it and succeeds because it's serving the clean pre-damage copy out of its cache.

Every other case in the suite asserts that the connector agrees with native, but here agreement means propagating a failure (since the file is meant to be authoritative). The connector has no invalidation signal for a modification made outside HDF5, e.g. nothing that could tell it the cached bytes no longer correspond to the file. Something like that is planned to be done in the intermediate future, but not immediately, so I opted to write the test and report it like this for now.


lukemartinlogan:
Hi, fix the merge conflict when you get a chance

lukemartinlogan (approved):

PR #779, opened 2026-07-20
Merge dev to main

lukemartinlogan:

PR #895, opened 2026-08-04
Fix owner-node cache blind spot behind the coherence CI failure

### LINKED ISSUE #894: Cache coherence 4-node suite fails on dev Cluster Tests since the #891 locality/perf merge
## Description
The 4-node cache coherence step passes on dev @ #887 merge but fails on dev @ #891 merge (run 30893447603): all four nodes exit 1, the visible tail shows the LAST scenario (repeated fragmented) dying on a Barrier timeout — a cascade signature from an earlier failure. The step takes ~26 min (barrier timeouts burning down) vs ~1 min locally.

The suite passes locally on the same code (repeated 5/5 on 4-node docker), so the failure is timing-dependent under the CI runner's 4-vCPU oversubscription (4 runtimes + 4 clients). Prime suspect: the known writer-local in-place-update vs foreign-invalidation race documented in the #891 commits (µs window locally; scheduling gaps widen it), or a timeout too tight for the constrained environment.

## Steps to Reproduce
`context-transfer-engine/test/integration/coherence/run_tests.sh` on a CPU-constrained docker host (mimic with --cpus limits).

## Acceptance Criteria
- [ ] Root cause identified (race vs timeout)
- [ ] Coherence step green on dev Cluster Tests

lukemartinlogan:
## Summary
- The blob-**owner** node's cache-slot copy sat outside the coherence protocol: origin registration skips self, so no foreign write ever invalidated it and the owner node stably served its own stale bytes whenever it was not the last writer — the round-order-dependent `repeated fragmented` winner split that failed dev Cluster Tests (and PR #779's check). Owner nodes now keep **no cache-slot copy**: their primary is already node-local DRAM and zero-IPC readable, so the copy bought no locality and doubled RAM.
- **Version-checked registration**: `GetBlob` reports the blob's content version in the OUT context; read-populate registrations carry it as `expected_version_` and the owner rejects stale ones (the reader deletes its local copy). Closes the fetch→register window that a racing put's invalidation snapshot cannot see.
- **Batch cache mirror is single-node only**: batched copies carry no registration, so on a cluster nothing would ever invalidate them.
- The coherence cluster config **disables SWIM**: on CI-class CPU budgets probe responses starve >30s and SWIM false-positives a healthy node dead — recovery reshuffles containers and every blob-barrier cascades to timeout (the second failure shape in the CI log, where a node was declared dead while its runtime demonstrably kept processing). The leader-election cluster test owns SWIM coverage; a genuinely dead node still fails this suite via barrier timeouts.
- Forensics for future flakes: `[COH]` runtime event logs (invalidate / register accept / self-refuse / version-reject), test-side winner decoding on mismatch, `COH_LOG_DIR` full-log dump in `run_tests.sh`, and a `docker-compose.cpulimit.yml` overlay (cpus: 0.5/node) that reproduces the CI timing locally.

## Motivation
Fixes #894 — dev Cluster Tests (and the dev→main PR check) red since the #891 merge.

Diagnosis trail: reproduced at `cpus: 0.5`; `[COH]` logs showed round-1 owner-serialization order 3,0,1,2 with writer 1 (the owner node) invalidating others but never registering itself, so winner 2's final put found an empty replica set and invalidated nothing. Node 1 then decoded its own `w1r1` pattern on every re-read while everyone else held `w2r1`.

## Test plan
- [x] 4-node coherence suite at CI-like CPU limits (cpus 0.5/node): passing, loop of 6 iterations in progress (iter 1 green; pre-fix the failure reproduced ~1 in 2)
- [x] Single-node CTE regression subset (`ctest -R 'cte_core_blob_put|cte_core_blob_get|cte_core_integration|cte_core_workflow|cte_functional_all|cte_core_tasks'`): 8/8 pass
- [ ] CI Cluster Tests green (note: the pre-existing #856 leader-election flake fails the same job independently)

## Breaking changes
None. Wire format: `Context` gains an OUT `version_` field and `RegisterReplicaContainerTask` an IN `expected_version_` (both ends ship together; no cross-version compatibility expected).

## Follow-ups
- A put targeting a SWIM-dead node can hang the submitting client forever (dead-node retry limbo) instead of erroring — observed while diagnosing; issue to follow.

### DIFF
--- a/context-transfer-engine/cache/include/clio_cte/cache/cache_runtime.h
+++ b/context-transfer-engine/cache/include/clio_cte/cache/cache_runtime.h
@@ -101,6 +101,16 @@ class Runtime : public clio::cte::core::CoreInterposer {
   void AimAtCacheReplica(Context *ctx) const;
 
   /**
+   * True when THIS node hosts the blob's owner container (issue #894).
+   * The owner node keeps NO cache-slot copy: its primary is already
+   * node-local DRAM (and zero-IPC readable), and an owner-side copy sits
+   * outside the register/invalidate protocol — origin registration skips
+   * self, so nothing would ever invalidate it and it serves stale bytes
+   * after a foreign write (the CI-caught fragmented winner split).
+   */
+  bool IsBlobOwnerLocal(const TagId &tag_id, const std::string &blob_name);
+
+  /**
    * Best-effort population of THIS node's raw local copy from the blob's
    * authoritative owner chain, plus coherence registration (the owner's
    * next primary write invalidates the copy). Sequential from 0, so an
--- a/context-transfer-engine/cache/src/cache_runtime.cc
+++ b/context-transfer-engine/cache/src/cache_runtime.cc
@@ -92,6 +92,31 @@ clio::run::PoolQuery Runtime::ScheduleTask(
   return CoreInterposer::ScheduleTask(task);
 }
 
+bool Runtime::IsBlobOwnerLocal(const TagId &tag_id,
+                               const std::string &blob_name) {
+  // Same hash as the core's HashBlobToContainer (owner routing), resolved
+  // against THIS pool's container map — every pool in the chain shares the
+  // container layout (replication's primary-hit path relies on the same
+  // property).
+  auto *pool_manager = CLIO_POOL_MANAGER;
+  const clio::run::PoolInfo *info = pool_manager->GetPoolInfo(pool_id_);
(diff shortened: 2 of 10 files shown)


lukemartinlogan:
Local validation complete: 6/6 iterations of the 4-node coherence suite green at cpus 0.5/node (the CI-timing repro that failed ~1 in 2 pre-fix). Single-node CTE regression subset 8/8.

Issue #886, opened 2026-08-01
Blob replicas in CTE + replication module: cache blobs in RAM with persistent replicas, no demotion tax


lukemartinlogan:
## Problem

CTE blobs are not replicable: a blob cached in the RAM tier must be *demoted* to a persistent tier to be durable — a performance tax (the hot copy is gone) and a general feature gap.

## Design

**CTE = mechanism only.** Blob metadata gains an indexed set of replicas; each replica is just its own block list:

- `struct Replica { priv::string name_; priv::vector<BlobBlock> blocks_; u64 total_size_cache_; }` — a label plus the block vector (same `BlobBlock` machinery as the primary: bdev client, target query, offset/size/capacity).
- `BlobInfo` gains `priv::vector<Replica> replicas_;` — empty (zero-cost) unless used; unbounded replica count.
- Replicas persist through the same metadata log as `blocks_`, so persistent-tier replicas survive restart.

**Addressing via Context (integer, as discussed):** `Context.replica_`:
- `0` (default): primary — today's behavior, bit-for-bit.
- `N > 0`: operate on replica N (lazy-created on first write; reads of an absent replica fail cleanly).
- `kAllReplicas (-1)`, writes only: write-through to primary + every existing replica.
Placed OUTSIDE the `#if CTP_ENABLE_COMPRESS` block (the transform_flags_ ABI lesson) and serialized unconditionally.

Tier selection for a replica write composes with the existing `persistence_target_` / `min_persistence_level_` fields — so "persistent replica of a RAM blob" is just `replica_=1, min_persistence_level_=2`. The SHM direct-read fast path serves the primary only (replica reads take the RPC path); Evict/Reorganize stay primary-only in v1 (documented).

**Replication module** (`context-transfer-engine/replication`, new chimod): owns the *policy/control* of keeping caches coherent between persistent and non-persistent modes. v1 API:
- `ReplicateBlob(tag, blob, replica_idx, ctx)` — runtime-side copy primary → replica (GetBlob → PutBlob with `replica_` set), tier chosen by ctx persistence fields.
- `FlushTag(tag, replica_idx, ctx, min_score)` — replicate all (matching) blobs of a tag; the "make this cache durable" sweep.
- Standard Create/Destroy/Monitor scaffolding; periodic/policy-driven replication can layer on later without CTE changes.

## Acceptance criteria
- [ ] Put/Get honor `Context.replica_` (specific replica read/write; `-1` write-through), primary path unchanged when `replica_==0`
[...]

Issue #862, opened 2026-07-29
Integrate Clio into YCSB: KV binding + benchmark comparison vs Redis, RocksDB, DynamoDB(-Local)


lukemartinlogan:
## Description

Add YCSB (Yahoo! Cloud Serving Benchmark) integration for Clio so the CTE blob store can be benchmarked with standard cloud-serving workloads and compared directly against Redis, RocksDB, and DynamoDB.

## Scope

1. **YCSB in `external/`**: clone `brianfrankcooper/YCSB` into `external/YCSB` (setup-provisioned like llama.cpp, not a submodule) and use the 0.17.0 release artifacts for the stock bindings.
2. **Clio YCSB binding**: a new `site.ycsb.db.ClioClient` DB adapter backed by a thin JNI shim over the CTE C++ client — key/value mapped to tag/blob (`insert`/`update` → PutBlob, `read` → GetBlob, `delete` → blob destroy; `scan` unsupported initially → run workloads A/B/C/F).
3. **Benchmark**: run YCSB core workloads (A: 50/50, B: 95/5, C: read-only, F: RMW) against Clio (SHM client + local runtime), Redis, RocksDB (rocksdbjni), and DynamoDB Local, on identical record/operation counts; publish throughput + latency comparison.
4. **Devcontainer**: augment the devcontainers to clone YCSB into `external/` by default and install the comparison services (redis-server, DynamoDB Local + Java runtime; rocksdb ships with the YCSB binding). Services installed and validated locally in the dev container first.

## Acceptance criteria
- [ ] `external/YCSB` provisioned with a runnable YCSB
- [ ] Clio binding loads/runs workloads A, B, C, F
- [ ] Comparison numbers for clio vs redis vs rocksdb vs dynamodb-local on the same workload parameters
- [ ] Devcontainer provisions YCSB + services out of the box

PR #906, opened 2026-08-04
Extract SemanticSearch into a new indexer chimod

### LINKED ISSUE #905: Extract SemanticSearch from CTE core into a new indexer chimod
## Description
`SemanticSearch` (Method 35) is currently implemented inside the CTE core runtime (`core_runtime.cc:7317-7525`) as a brute-force scan: every query regex-filters tags/blobs, reads the full bytes of every candidate blob off the bdev, tokenizes, and computes BM25 on the fly. The API is still necessary, but reverse-index management is not a core responsibility — core's contract is authoritative content storage/transfer, while a search index is derived, rebuildable state with its own storage/compute/consistency profile.

This issue extracts SemanticSearch into a new **indexer chimod** (`context-transfer-engine/indexer`) that:
- Sits in the compose chain as an interposer (deriving from `CoreInterposer`) with `next_pool_id` pointing at CTE core, forwarding all non-search methods down the chain (same pattern as replication/cache).
- Keeps **Method 35 with core's existing `SemanticSearchTask` struct**, so all existing consumers (CAE forward, CEE, Python bindings, `cte_search` CLI, benchmark) keep working unchanged.
- Maintains a real **inverted index** (term → postings with term frequencies), updated incrementally by intercepting PutBlob/DelBlob/DelTag/TruncateBlob forwards, instead of re-reading every blob on every query.
- **Rebuilds the index from storage after a runtime reboot** (driven by the restart lifecycle, mirroring `RebuildTagSearchIndexLocked`), so search results are correct post-restart.

`TemporalSearch` (Method 36) stays in core for now — it is a pure metadata scan over core-internal maps; moving it would require a new metadata enumeration API. Possible follow-up.

## Acceptance Criteria
- [ ] New chimod `context-transfer-engine/indexer` (pool id 564.0) with client/runtime/tasks/lib_exec, registered in the CTE CMake tree behind `CLIO_CTE_ENABLE_INDEXER`
- [ ] Indexer handles `kSemanticSearch` from its inverted index; CTE core's runtime handler is removed
- [ ] Index updated incrementally on PutBlob/DelBlob/DelTag/TruncateBlob interception
- [ ] Index rebuilt correctly from persisted storage on restart; new restart test proves search returns correct results after reboot
- [ ] All existing SemanticSearch tests pass (`python_semantic_search_test`, `cae_semantic_search`, `cae_labeling_dispatch`, CEE query tests, `cr_autogen_sweep_tests`) with at most compose-config updates
- [ ] Default compose chain documentation updated (`clio_default.yaml`, configuration docs)

lukemartinlogan:
## Summary
- New `context-transfer-engine/indexer` chimod (pool 564.0): an interposer above the CTE core that owns the SemanticSearch index. It maintains a per-container forward index (term frequencies + doc lengths) by intercepting PutBlob/MultiPutBlob/TruncateBlob (forward-then-re-read) and DelBlob/DelTag/RenameTag, serves Method 35 from that index with the core's exact BM25 semantics, and forwards every other core method down `next_pool_id` (cache/replication pattern, `CoreInterposer` base).
- The CTE core no longer executes SemanticSearch — no more re-reading every candidate blob's bytes per query. The task struct, method id, and client API stay in core so every consumer (CAE forward, CEE, Python bindings, `cte_search`, benchmark) works unchanged; a Method-35 task reaching a core with no indexer composed now returns an explicit error instead of a silent scan.
- **Restart**: the index is derived state, never persisted. On the compose restart path each container rebuilds its slice from the storage below (BlobQuery + GetOrCreateTag + GetBlob) before Create acks. New `cte_indexer_restart` integration test (file-backed tier, two daemons, two-phase client) proves post-reboot searches return byte-identical BM25 rankings with zero puts after the restart.
- **Two latent core restart bugs fixed** (exposed by the rebuild test): `RestoreMetadataFromLog` never resynced `total_size_cache_` after rebuilding `blocks_`, so every restored blob reported size 0 (a Debug build asserts in `GetTotalSize`); and `kCreateNewBlob` WAL replay reset restored block layouts — the third instance of the outlives-the-flush hazard, alongside transform flags (#818) and replicas (#886).
- Default chain updated: `cache(563) -> indexer(564) -> [compressor(562) ->] replication(561) -> core(512)` — indexer sits above the compressor so its read-backs see logical bytes.

## Motivation
Closes #905. Search-index management is derived-data policy, not core storage responsibility; core's contract stays put/get/metadata, and the index becomes rebuildable layer state.

## Test plan
- [x] New restart test passes: `ctest -R cte_indexer_restart` (rankings identical pre/post reboot)
- [x] Restart/WAL regressions on the core fixes: `cr_cli_cte_wal_restart`, `cr_cli_compose_restart`, `cr_cli_restart_log`, `cte_restart_integration`, `cte_replication_persist_integration` all pass
- [x] `python_semantic_search_test`, `python_temporal_search_test`, `python_bindings_test`, `python_types_test` pass (client bound to the indexer via `CLIO_CTE_POOL=564.0`)
- [x] `cae_labeling_dispatch` passes (CAE → indexer SemanticSearch forward)
- [x] CEE suite 19/19 including `cee_query_prompt_bm25` (CEE → CAE → indexer chain; fixture now creates the CAE pool with `next_pool_id`=indexer *before* `CLIO_CAE_CLIENT_INIT`, whose default create used to win)
- [x] `cr_autogen_sweep_tests` passes (core keeps Method-35 serialization cases)
- [ ] `cae_semantic_search` needs live Ollama — not runnable on this box, CI covers it
- [x] cpplint clean on all new files; BSD headers applied

## Breaking changes
None for API consumers (method id + task structs unchanged). Deployments composing the core WITHOUT the indexer now get an explicit error (rc=2 + log) on SemanticSearch instead of results — compose `clio_cte_indexer` above the core (the default config does). Follow-ups: TemporalSearch (Method 36) intentionally stays in core (pure metadata scan); docs-site configuration page needs the new chain diagram (docs repo).

### DIFF
--- a/context-assimilation-engine/test/unit/test_cae_labeling_dispatch_config.yaml
+++ b/context-assimilation-engine/test/unit/test_cae_labeling_dispatch_config.yaml
@@ -21,12 +21,14 @@ compose:
     bdev_type: ram
     capacity: "256MB"
 
-  # CAE core at the entrypoint (512.0), labeling enabled.
+  # CAE core at the entrypoint (512.0), labeling enabled. Forwards to the
+  # indexer (564.0), which owns SemanticSearch (issue #905) over the CTE
+  # core at 513.0.
   - mod_name: clio_cae_core
     pool_name: cae_label
     pool_query: local
     pool_id: "512.0"
-    next_pool_id: "513.0"
+    next_pool_id: "564.0"
     label_endpoint: "http://127.0.0.1:1"
     label_prompts:
       summarize: "Summarize this document:"
@@ -56,3 +58,11 @@ compose:
         score: 1.0
     dpe:
       dpe_type: "max_bw"
+
+  # Indexer chimod (issue #905): owns SemanticSearch; the CAE forward
+  # terminates here instead of the core.
+  - mod_name: clio_cte_indexer
+    pool_name: clio_cte_indexer
+    pool_query: local
+    pool_id: "564.0"
+    next_pool_id: "513.0"
--- a/context-assimilation-engine/test/unit/test_cae_semantic_search_config.yaml
+++ b/context-assimilation-engine/test/unit/test_cae_semantic_search_config.yaml
@@ -1,7 +1,7 @@
 # Compose configuration for the CAE labeling + CTE semantic-search test.
 # Same shape as the labeling test: CAE intercepts at 512.0 with one
-# labeling rule that fires on every blob; CTE backs it at 513.0 and
-# owns the BM25 search.
+# labeling rule that fires on every blob; the indexer chimod at 564.0
+# owns the BM25 search (issue #905) over the CTE core at 513.0.
 
 networking:
   port: 9413
@@ -22,7 +22,7 @@ compose:
     pool_name: cae_main
     pool_query: local
     pool_id: "512.0"
-    next_pool_id: "513.0"
+    next_pool_id: "564.0"
     label_endpoint: "http://127.0.0.1:11434"
     label_prompts:
       summarize: "Summarize the following text in one short sentence."
@@ -45,3 +45,11 @@ compose:
         score: 1.0
     dpe:
       dpe_type: "max_bw"
+
+  # Indexer chimod (issue #905): CAE forwards PutBlob and SemanticSearch
+  # here; puts maintain the index inline and the search is served from it.
+  - mod_name: clio_cte_indexer
+    pool_name: clio_cte_indexer
+    pool_query: local
+    pool_id: "564.0"
+    next_pool_id: "513.0"
--- a/context-exploration-engine/api/test/CMakeLists.txt
+++ b/context-exploration-engine/api/test/CMakeLists.txt
@@ -12,10 +12,12 @@ function(add_cee_api_unit_test TEST_NAME TEST_SOURCE)
   # Create test executable
   add_executable(${TEST_NAME} ${TEST_SOURCE})
 
-  # Link common dependencies
+  # Link common dependencies (indexer client: the SemanticSearch pool the
+  # BM25 tests compose over the core, issue #905)
   target_link_libraries(${TEST_NAME}
     PRIVATE
       clio_cee_api
+      clio_cte_indexer_client
   )
 
   # Install test binary
@@ -52,6 +54,7 @@ add_executable(test_context_comprehensive test_context_comprehensive.cc)
 target_link_libraries(test_context_comprehensive
   PRIVATE
     clio_cee_api
+    clio_cte_indexer_client
 )
 install(TARGETS test_context_comprehensive DESTINATION bin/tests)
 
--- a/context-exploration-engine/api/test/test_context_comprehensive.cc
+++ b/context-exploration-engine/api/test/test_context_comprehensive.cc
@@ -51,6 +51,7 @@
 #include <clio_cae/core/constants.h>
 #include <clio_cae/core/factory/assimilation_ctx.h>
 #include <clio_cte/core/core_client.h>
+#include <clio_cte/indexer/indexer_client.h>
 #include <clio_runtime/bdev/bdev_tasks.h>
 #include <clio_runtime/clio_runtime.h>
 #include <fstream>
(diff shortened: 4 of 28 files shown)


lukemartinlogan:
## Performance A/B (indexer composed vs not — same binaries, compose-only difference)

Box: WSL2, 6 runtime workers, RAM tier, medians of 3 interleaved iterations (2 for lmcache). **A** = client→core direct; **B** = client→indexer(564)→core, default scope (`.*`, everything indexed); **C** (fio only) = indexer composed but scoped out (`tag_re: "^$"`).

### clio_cte_bench (4 threads, depth 8, 10s timed, bounded keyspace)

| op | A | B | delta |
|---|---|---|---|
| Put 4k | 38.2k IOPS | 31.7k IOPS | −17% |
| PutDefer 4k | 8.4k IOPS | 7.0k IOPS | −17% |
| Put 1m | 8,100 MB/s | 3,640 MB/s | −55%* |
| PutDefer 1m | 3,883 MB/s | 2,212 MB/s | −43%* |
| Get 4k | 2.47M IOPS | 2.30M IOPS | −7% |
| Get 1m | 8,041 MB/s | 7,983 MB/s | ≈0 |
| GetDefer 4k / 1m | 2.59M / 7,969 | 2.45M / 7,584 | −5% |

\* sustained-saturation tax: the bench overwrites hot keys at max rate for 10s, so the async sweep is continuously re-reading+re-tokenizing. The measured passive floor (`CLIO_INDEXER_PASSIVE=1`, forward-only) is ~5.5 GB/s / ~24k IOPS — bursty or read-heavy workloads ack there. Before the async-indexing commit this row read **347 MB/s (−95%)**.

### fio over clio-fs (psync, 4 jobs, --thread; 4k over 32m/job, 1m over 256m/job)

| job | A | B (in scope) | C (scoped out) |
|---|---|---|---|
| 4k write | 52.8 MiB/s | 62.3 | 47.5 — all noise (cfs client-deferred writes absorb the chain) |
| 4k randread | 5,565 MiB/s | 5,120 | 5,333 — noise |
| 1m write | 814 MiB/s | 720 | 736 — ~−10% forward hop |
| 1m read | 16.7 GiB/s | 9.9 (−41%) | 15.9 — B's dip = sweep still digesting the just-written 1GB of in-scope pages |

### lmcache store (embedded runtime, CLIOKV1 records via the defer registry, 4 submit threads)

| workload | A write / read | B write / read |
|---|---|---|
| 256 × 3.7MB | 208 / 300 ops/s | 202 / 268 ops/s (≈parity) |
| 64 × 14.7MB | 50 / 41 ops/s | 62 / 39 ops/s (≈parity) |

### Takeaways
- **Chains without the indexer are byte-identical to before this PR** — zero impact.
- With the indexer composed, **reads are free** (no interception, just the scheduling hop) and **puts pay only an O(1) coalesced enqueue** — the visible costs are the interposition dispatch at small sizes (~15-20%) and background sweep competition when a write-saturated in-scope working set is being digested.
- **Binary-data workloads (fio pages, KV tensors) should be scoped out** via `tag_re`/`blob_re` — side C shows scoped-out cost is pure interposition (noise on cfs).
- Put-path history for the record: the first implementation cost −95% at 1 MiB. Three fixes got it to −55%-at-saturation/≈0-at-burst: client-API round-trips replaced with nested inline runs on the core container (the `CLIO_RUN_INLINE` #862 cost profile), the per-put read-back removed, and the per-byte locale tokenizer (`std::isalnum` per byte ≈ 350 MB/s ceiling) tabled and moved off the ack path entirely.

Also in this A/B cycle: `clio_cte_bench` gained `PutDefer`/`GetDefer`/`PutGetDefer` cases (Defer API coverage), and the WAL-persistence commit means restarts restore the index instead of rescanning (restart test now proves phase-2 searches from the restored WAL with zero puts and zero rescans).

lukemartinlogan:
CI note: the `adapters (macos, hdf5_vol + macfuse)` failure is **pre-existing on dev**, not introduced here — dev tip `2992817c` shows the same leg cancelled at the 2h timeout (twice), with the identical admin-pool `RouteTask: RouteLocal returned 4` retry storm (the #841/#849 startup-wedge class). Filed as #914; the leg has been rerun on this PR. Every other leg is green and codecov/patch passes.

Issue #848, opened 2026-07-28
Windows CI: one-different-test-per-run flake cluster; umap_ll lost-key on windows-11-arm indicates a real weak-memory race


lukemartinlogan:
## Symptom

Every Windows CI run this morning (2026-07-28) fails exactly **one** test — and it is a **different test every run**, across unrelated branches:

| Run | Branch | Job | Failing test |
|---|---|---|---|
| 30342597188 | 841-route-retry-null-worker | build-test (windows-11-arm) | `ctp_priv_umap_ll` — "concurrent inserts survive concurrent growth", `find()` returned nullptr for one of 48k keys after joins |
| 30342597188 | 841-route-retry-null-worker | icx (windows-2022) | `cte_bdev_fragmentation` — `large_ok == kNumLarge` |
| 30342046100 | fix-routetask-nonworker-retry-drop | icx (windows-2025-vs2026) | `cr_detached_spawn` |
| 30341734171 | 727-main-segment-ram-default | build-test (windows-2025) | `cte_tag_force_net` (Timeout) |
| 30324753001 | 822-spsc-task-queue | icx ×2 + build-test | `ctp_ring_buffer_ext` (Timeout), `cr_detached_spawn` ×2 |

The same jobs passed an hour earlier on dev (30340004985 = success) and on PR #839. One-failure-per-run with no repeat suggests scheduling/timing sensitivity on the Windows runners (2-vCPU oversubscription) rather than one specific regression.

## The umap_ll ARM failure deserves a real look

`ctp_priv_umap_ll` on **windows-11-arm** (MSVC ARM64, weak memory model) lost an inserted key: the test's `size() == 48000` check passed but `find(k)` returned nullptr — i.e. `size_` counted an insert whose node is not reachable from any bucket, which is the signature of a mutual-exclusion violation between an insert and a concurrent rehash (rehash re-threading from a stale bucket head would orphan a just-linked node).

Analysis done so far (didn't find the hole):
- On host, `ipc::atomic` = `std_atomic` forwarding to `std::atomic` with seq_cst defaults everywhere — no relaxed/volatile shortcuts in the host path.
- The `RwLock` protocol (mode_/readers_/writers_/ticket_/cur_writer_) checks out under seq_cst: a reader can only return while registered AND mode==kRead was observed, and mode cannot leave kWrite while a writer is registered, so reader/writer co-residency should be impossible; writer/writer is serialized by the ticket.
[...]

PR #917, opened 2026-08-05
Survive conda-build's transient internal crash in the dependency install

### LINKED ISSUE #916: CI: intermittent conda-build 26.7.0 IndexError wedges CI Linux dependency install
## Description
`CI/ci-deps.sh` intermittently dies during `conda build` with an internal conda-build crash — not our recipe, and not our code:

```
Attempting to finalize metadata for iowarp-core
Getting pinned dependencies: ...working... done
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<
    Traceback (most recent call last):
      ...
      File ".../conda_build/render.py", line 1034, in expand_outputs
        for output_dict, m in deepcopy(_m).get_output_metadata_set(
    IndexError: list index out of range

    conda version : 26.7.0
    conda-build version : 26.7.0
```

## Evidence that it is intermittent, not a code regression
Two CI Linux runs fired on the **same dev SHA** `c95cecdba` at 04:26 on 2026-08-05:

| run | build-test (prediction/cae/cee) + leak-check |
|---|---|
| [30975138393](https://github.com/iowarp/clio-core/actions/runs/30975138393) | all four **failed** at the ci-deps step |
| [30975136879](https://github.com/iowarp/clio-core/actions/runs/30975136879) | all four **passed** |

Same commit, same workflow, concurrent runs, opposite outcomes — so the trigger is environmental (conda tooling/channel state), not the tree. Note the runner ships conda 26.5.3 and the step ends up on 26.7.0, so a version bump is in play; this is the same class of breakage as #874 (runner image broke conda-build; fixed by pinning Miniconda).

## Impact
When it hits, four CI Linux jobs go red for reasons unrelated to the change under test, which costs triage time on every PR that draws the bad draw.

## Acceptance Criteria
- [ ] Pin conda/conda-build to a known-good version in `CI/ci-deps.sh` (or the Miniconda install step), mirroring the #874/#875 pin approach
- [ ] Confirm the pinned version does not reproduce the IndexError across several consecutive runs

lukemartinlogan:
## Summary
- Retry the `conda build` call in `CI/ci-deps.sh` **only** when the failure carries conda-build's internal-crash signature (`ERROR REPORT` / `IndexError: list index out of range`), so genuine build failures still fail on the first attempt rather than costing a second 10–30 minute build.
- Widen the `conda render` retry backoff from 10/20/30s to 30/60/120s.

## Motivation
Fixes #916. On dev `c95cecdba` four CI Linux jobs (prediction/cae/cee/leak-check) and both macOS legs failed in the dependency-install step with a crash raised *inside conda-build itself*:

```
File ".../conda_build/render.py", line 1034, in expand_outputs
IndexError: list index out of range
```

Evidence that it is transient rather than a code or version problem:
- **8 of 12 jobs in the same failing run passed the identical step** — `build-test (unit amd64)` ran the same `CI/ci-deps.sh` with conda-build on the same commit at the same minute and succeeded.
- **The same four jobs passed in the sibling run** on that identical SHA.
- **Passing and failing jobs both ran conda-build 26.7.0**, so a version pin would not have helped.

`ci-deps.sh` already retries `conda create`, the conda-build install, and `conda render` — the full-build path was the one gap, which is exactly where the Linux jobs died.

## Test plan
- [x] `bash -n` clean
- [x] Behavior verified against a stubbed conda binary in four scenarios: success → 1 attempt; **real build error → 1 attempt, no retry** (the important one); internal crash then recovery → 2 attempts, succeeds; persistent crash → stops at 2
- [ ] CI green on this PR (exercises the modified script on every Linux/macOS job)

## Breaking changes
None. Behavior is unchanged except that a conda-build internal crash now gets one retry.

### DIFF
--- a/CI/ci-deps.sh
+++ b/CI/ci-deps.sh
@@ -352,6 +352,10 @@ if [ "$ONLY_DEPS" = true ]; then
     # pin and trips conda-build's execute_download_actions IndexError.
     # Retry: render performs a solve and can hit the same transient
     # "database is locked" / conda-forge 502 flakes as the installs above.
+    # Backoff is 30/60/120s (not 10/20/30): on 2026-08-05 a render episode
+    # outlived all three of the shorter waits and failed the macOS legs
+    # (#916), while sibling jobs on the same commit rendered fine minutes
+    # later. The wider window costs nothing on the happy path.
     _render_ok=0
     for _attempt in 1 2 3; do
         if "$CONDA_BIN" render "$RECIPE_DIR" \
@@ -360,8 +364,9 @@ if [ "$ONLY_DEPS" = true ]; then
             _render_ok=1
             break
         fi
-        echo -e "${YELLOW}conda render failed (attempt $_attempt/3), retrying in $((10 * _attempt))s...${NC}"
-        sleep $((10 * _attempt))
+        _render_wait=$((30 * (1 << (_attempt - 1))))
+        echo -e "${YELLOW}conda render failed (attempt $_attempt/3), retrying in ${_render_wait}s...${NC}"
+        sleep "$_render_wait"
     done
     if [ "$_render_ok" -ne 1 ]; then
         echo -e "${RED}conda render failed${NC}"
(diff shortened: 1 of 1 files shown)


lukemartinlogan:
All 34 checks green. Both legs that exercise the change are covered: the Linux `build-test (prediction/cae/cee amd64)` + `leak-check` jobs and the macOS `build-test`/`boost`/`adapters` jobs all pass through the modified `CI/ci-deps.sh` conda path.

Two unrelated infrastructure failures showed up during validation and both cleared on re-run, worth noting so they are not mistaken for this change:
- `build-test (windows-11-arm)`: the #848 flake (`cr_safe_bdev_concurrency_tests` startup wedge). Windows never executes `ci-deps.sh` — it uses vcpkg — so the change cannot reach it.
- `build-test (macos-26-intel)`: GitHub lost the runner 70 minutes in, mid-`ctest`, with the log blob never uploaded (`BlobNotFound`) against a 120-minute job timeout. Its `Install dependencies (conda)` step had already succeeded.

Ready for review.

Issue #915, opened 2026-08-05
AggregateOut must not delegate to Copy(): whole-task assignment corrupts the origin (77 instances tree-wide)


lukemartinlogan:
## Description

`Task::AggregateOut` documents an explicit contract:

> IMPORTANT: Derived classes that override AggregateOut MUST call `Task::AggregateOut(replica_task)` first before aggregating their own fields.

`AggregateOut` merges a **replica's OUT fields** into the origin (N→1 gather). But **77 task types across the tree** implement it as:

```cpp
void AggregateOut(const FullPtr<Task> &other_base) {
  Task::AggregateOut(other_base);
  Copy(other_base.template Cast<XTask>());   // <-- whole-task assignment
}
```

`Copy()` is a *whole-task* assignment, which does two illegal things during aggregation:

1. **It runs `Task::Copy`**, overwriting the ORIGIN's `task_id_`, `pool_query_`, `method_`, `task_flags_`, `completer_` and `task_group_` with the REPLICA's — destroying the origin's identity while `send_map_`, the replica accounting and the completion path still reference it. The replica's `task_id_` carries `net_key`/`replica_id`, so the origin adopts a subtask's identity mid-flight.
2. **It re-assigns IN fields**, including `priv::string` / shm-allocated members, from the replica's shared-memory segment — freeing the origin's buffer through the wrong allocator.

Result: `free(): invalid pointer` / SIGSEGV (exit 139), plus silently wrong OUT values (last-replica-wins instead of a real merge).

## Confirmed live impact (issue #856)

Reproduced locally in ~2 minutes with `context-runtime/test/integration/leader_elect/run_tests.sh all`:

1. Test kills the leader.
2. New leader runs `TriggerRecovery` → broadcasts `RecoverContainers` → aggregates replies → **`free(): invalid pointer`, exit 139**.
3. Next leader takes over, same crash, exit 139.
4. Third node sees 2/3 peers dead and **self-fences** (`Recovery: Skipping for node N - self-fenced`), so recovery NEVER runs.
5. Containers stay mapped to dead nodes; the client's post-failover `CreatePool` hangs forever → the 15-minute CI step timeout that has been flaking Cluster Tests.

## Broadcast-reachable instances (fixed on branch `856-leader-election-failover`)

These are the ones where aggregation actually runs with a foreign replica today:

[...]

Issue #896, opened 2026-08-04
Put to a SWIM-dead node hangs the submitting client forever instead of erroring


lukemartinlogan:
## Description
Observed while diagnosing #894 (coherence suite at cpus 0.5/node): when SWIM marks a peer dead, `SendIn` skips sends to it ("target node N is dead ... skip" / "queuing for retry") and `ScanSendMapTimeouts` logs "timed out waiting for dead node" — but the origin client's `put.Wait()` never returns (observed >20 minutes). The task neither completes nor errors back to the submitter, so application threads hang indefinitely.

## Steps to Reproduce
1. 4-node coherence cluster (`context-transfer-engine/test/integration/coherence`) with the cpulimit overlay (cpus: 0.5/node) and SWIM enabled with default timeouts.
2. Under cold-start CPU starvation, SWIM false-positives peers dead within the first minute.
3. The test's first `AsyncPutBlob` whose owner container hashes to a "dead" node hangs forever; the container never exits.

## Expected vs Actual
- **Expected:** the task completes with a network-timeout RC (#628 family) so the client can fail fast, as `HandleTaskProgressResult` already does for in-flight tasks whose replica is Gone.
- **Actual:** `ScanSendMapTimeouts` logs the timeout but the origin future is never completed; the client blocks forever.

## Acceptance Criteria
- [ ] A task addressed to a node in `kDead` state completes its origin future with the network-timeout RC within the retry deadline.
- [ ] Regression test: mark a node dead, submit a put routed to it, assert `Wait()` returns nonzero within the deadline.

lukemartinlogan:
**First CI manifestation, in the Cluster Tests reconnect suite** (dev `c95cecdba`, 2026-08-05 04:41): [run 30975136893](https://github.com/iowarp/clio-core/actions/runs/30975136893).

`test_reconnect.cc` step 3 does `AsyncCreate(PoolQuery::Dynamic(), "reconnect_post_failover_pool")` right after step 2 kills the local runtime. The client failover itself works — the log ends cleanly with:

```
WaitForServerAndReconnect: Trying 1/16: 172.26.0.13
ReconnectToNewHost: Connected to 172.26.0.13 (generation=862146218065)
```

…and then produces **zero further output for the full 15-minute step timeout**. Same shape as this issue: an operation whose coordination must involve the just-killed node neither completes nor errors back, so the submitting client blocks forever rather than failing fast.

[...]

PR #918, opened 2026-08-05
Leader election & failover: design, dead-node completion, and the recovery data race (#856)

lukemartinlogan:
> **Draft.** Fixes the `leader_elect` CI failure that was blocking dev→main. Verified green in CI; two non-blocking check failures explained at the bottom.

## Status

**Cluster Tests is GREEN in CI on this branch** — run [31005222978](https://github.com/iowarp/clio-core/actions/runs/31005222978):

```
success   Runtime leader-election test (leader death + re-election)
success   CTE cache coherence test (4-node, issue #886)
```

Locally, `leader_elect` Phase 1 runs **3/3, zero crashes**. Full local `ctest`: **293/302**, where all 9 failures are docker wrappers hitting a WSL path-mount quirk (bare ctest doesn't set `HOST_WORKSPACE`) — **zero code failures**.

| | Before | After |
|---|---|---|
| Node crashes | every run, exit 139 (heap corruption) | **0** |
| `leader_elect` Phase 1 | never passed; CI step hit its 15-min timeout | **3/3 pass** |

⚠️ **`cluster-tests.yml` never runs on branch PRs** — only on `push` to main/dev plus `workflow_dispatch`. A branch can look entirely green while the suite it fixes was never executed. I dispatched it manually; anyone touching cluster behaviour should do the same.

## Why one red step took six fixes

The symptom **moved with build layout** — `free(): invalid pointer` → `double free detected in tcache 2` → `Fatal glibc error: tpp.c:86 __pthread_tpp_change_priority` inside `std::mutex::unlock()` → silent SIGSEGV. Each move looked like a new bug. Seven theories were eliminated with evidence before the real causes surfaced (all recorded in #856 so nobody re-chases them).

1. **`recovery_initiated_` data race.** `HeartbeatProbe` is periodic and runs on *different worker threads* across periods; with two nodes down, two `TriggerRecovery` calls mutate a plain `unordered_set` concurrently and the rehash (bucket_count **1 → 13**) frees the bucket array under the other thread. Now a process-wide atomic claim — deliberately *not* per-container, since recovery re-registers admin containers while probes are running on them.
2. **Dead-node tasks hung forever (#896).** The dead-node sweep only examined `Physical`-mode pool queries, but `Dynamic`/`Broadcast` resolve to `DirectId`/`Range` whose node is looked up at dispatch time — and the probe path skips dead targets, deferring to "the dead-node timeout path" that could not see them. Nobody completed those origins. Now driven off `progress_map_`.
3. **`AggregateOut` delegating to `Copy()`** in 4 broadcast-reachable tasks — a whole-task assignment that overwrote the origin's `task_id_`/`pool_query_`/`completer_` with a replica's and re-assigned shm strings across segments. **~73 latent instances remain tree-wide, including the `MOD_NAME` module template → #915.**
4. **Admin pool redistributed during recovery** — per-node control-plane containers moved onto survivors that already had one.
5. **Pool creation failed when a node died**, because one undeliverable replica poisoned the origin's return code. Forgiven in `BaseCreateTask::PostWait` — it *must* live there, since `HandleTaskProgressResult` applies the dead-replica verdict after the survivors aggregate.
6. **SWIM timings in the suite** (2s/2s/3s) false-positived *healthy* nodes under docker load, after which survivors self-fenced and the create never completed. Relaxed to 5s/5s/8s — safe now that creates tolerate an undeliverable replica.

## ASan works on fibers now

`__sanitizer_start/finish_switch_fiber` annotations around the Boost.Context switches, enabled automatically with `CLIO_CORE_ENABLE_ASAN`. Previously ASan died with `DEADLYSIGNAL / nested bug in the same thread` on the first fiber fault, so it could diagnose **nothing** in a runtime where every task body runs on a fiber. It found no memory error here; its value was removing crash noise so bugs 5 and 6 became legible.

## Reverted — do not retry

Sizing pools to the live cluster. `CreatePool` runs on *every* node handling the broadcast, so each computed its own container count from its own transient liveness view: a restarted node built a pool with 4 containers while survivors built 3. Pool geometry must stay globally consistent; both sites now carry comments saying why liveness must not be consulted there.

## Deliberately out of scope (tracked, not claimed fixed)

- **Phase 2 / node rejoin** — `[leader_elect][post_restart]` hangs creating a pool whose id Phase 1 already created: survivors log "already exists by ID" while the restarted node, having lost its state, tries to create it fresh. Node-rejoin state restoration is a feature gap. **Pre-existing** — before this branch it failed even earlier, unable to initialize a client at all.
- **SWIM treats CPU starvation as death**, and self-fencing then disables recovery outright on small clusters. The timing change above is a test-side mitigation only. Phase 4 of `context-runtime/docs/leader_election_failover.md`.
- **#896** (dead-node client hang), **#915** (~73 `AggregateOut` instances).

## Non-blocking check failures

- **`build-test (windows-2025)`** — `cte_bdev_leak_stress` **Timeout**; 226/227 Windows tests pass. Known flake that also fails on dev, with prior commit history settling this same test. Unrelated to anything here.
- **`codecov/patch`** — coverage on new lines. Most of this branch is distributed recovery code that only executes in multi-node docker clusters, which do not feed coverage. I added a single-node regression guard for the `AggregateOut` contract (origin identity survives aggregation; `num_recovered_` sums instead of being clobbered — both fail against the old implementation), but it does not clear the threshold alone.

## Test plan

- [x] `leader_elect` Phase 1: 3/3 local, zero crashes
- [x] Cluster Tests green in CI (manually dispatched)
- [x] Full local `ctest`: 293/302, zero code failures
- [x] Regression subset 14/14 after every behavioural change
- [ ] Phase 2 / node rejoin (out of scope, tracked above)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -152,3 +152,8 @@ build-tsan/
 
 # Local-only compose overrides (per-box mounts/debug shims) must never be committed
 docker-compose.override.yml
+
+# Local-only ASan/backtrace debug shim (issue #856) — built ad hoc inside the
+# test containers, never part of the build.
+abrt_bt.c
+abrt_bt.so
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1131,6 +1131,12 @@ if(NOT CLIO_CORE_MSVC_LIKE)
         message(STATUS "AddressSanitizer + LeakSanitizer (ASan/LSan) enabled for iowarp-core")
         add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
         add_link_options(-fsanitize=address)
+        # issue #856: annotate Boost.Context stack switches. Without these ASan
+        # keeps shadow state for the stack it believes is current, so the first
+        # fault on a fiber stack aborts with "DEADLYSIGNAL / nested bug in the
+        # same thread" and reports NOTHING — which makes ASan useless across
+        # most of the runtime, where tasks run on fibers.
+        add_compile_definitions(CLIO_ASAN_FIBERS)
     endif()
 
     # UndefinedBehaviorSanitizer configuration
--- a/context-runtime/docs/leader_election_failover.md
+++ b/context-runtime/docs/leader_election_failover.md
@@ -1,0 +1,183 @@
+# Leader Election & Failover: Design (issue #856)
+
+Status: DRAFT for review · Branch: `856-leader-election-failover` · 2026-08-04
+
+## 1. What the leader is for
+
+Clio's "leader election" is not consensus. Every node independently computes
+
+```
+leader = lowest alive node_id            (ipc_manager.cc GetLeaderNodeId)
+```
+
+from its **local** SWIM membership view. The leader's purposes, in order of
+importance:
+
+1. **Exactly-once failure recovery.** When SWIM confirms a node dead, only the
+   leader runs `TriggerRecovery` (admin_runtime.cc): it redistributes the dead
+   node's containers and broadcasts the new container→node mappings. The
+   single-coordinator rule is what prevents N survivors from concurrently
+   remapping the same containers.
+2. **Collective aggregation rendezvous.** Neighborhood leaders (lowest alive
+   id within a neighborhood) are the batch points for ManyToOne collectives.
+
+Everything else (pool creation, routing, barriers) is leader-agnostic.
+
+## 2. Current failure evidence (dev, 2026-08-04, 3 consecutive CI timeouts)
+
+The historical #856 crash (fiber frame corruption in `TriggerRecovery`) was
+fixed in July (PR #878: exactly-once origin completion + strict event-resume
+guard). The step now fails differently. From the attempt-3 log of run
+30910530528:
+
+```
+Step 2: kill local runtime (the leader, node 0)
+Step 3: Creating pool on new host after reconnection
+  Phase 1: ReconnectToOriginalHost x5 (shm_attach ENOENT) — expected, 5s budget
+  Phase 2: Trying 16 random hosts
+    172.24.0.10 not responding            (the dead leader — expected)
+    Connected to 172.24.0.11 (generation=1040199788761)   ← FAILOVER WORKED
+  <silence for 15 minutes — step timeout>
+```
+
+**The client-side transport failover succeeds.** The hang is the FIRST
+operation submitted over the new connection: a `PoolQuery::Dynamic()`
+`CreatePool`, whose container fan-out includes the dead node. That request
+never completes and never errors. Three protocol gaps compound here:
+
+- **Gap A — tasks addressed to dead nodes hang forever (issue #896).** Before
+  SWIM confirms the death (30s direct-probe + suspicion), `SendIn` to the dead
+  node waits/retries; after `SetDead`, sends are skipped and
+  `ScanSendMapTimeouts` logs "timed out waiting for dead node" — but the
+  origin future is never completed. A broadcast with one replica on a dead
+  node therefore never finishes. The 15-minute timeout proves this is a
+  permanent hang, not slow detection.
+- **Gap B — no epoch/fencing on membership-derived decisions.** "Leader" and
+  "alive" are per-node opinions. During churn, two nodes can act as leader
+  simultaneously (double recovery) or a stale leader can keep coordinating
+  after the survivors moved on. Recovery, container remaps, and collectives
+  carry no membership version, so stale actors cannot be detected or refused.
+- **Gap C — failure detection is starvation-fragile.** SWIM probe responses
+  ride ordinary admin tasks; on an oversubscribed host a >30s CPU stall reads
+  as death (observed in #894: a demonstrably-alive node was declared dead,
+  recovery reshuffled its containers, every barrier cascaded). False-positive
+  death is *destructive* — recovery moves containers — so the cost asymmetry
+  demands a high-confidence signal.
+
+## 3. Design goals
+
+1. A task whose target is (or becomes) dead completes with a network-timeout
+   RC within a bounded time. No caller ever hangs on a dead node.
+2. Recovery is exactly-once **per membership change**, enforceable even under
+   divergent views: a stale coordinator's actions are refused, not raced.
+3. Failure detection distinguishes "dead" from "starved" cheaply enough for
+   CI-class CPU budgets.
+4. The existing deterministic min-alive-id rule stays. It is simple, has no
+   election traffic, and is fine *once views agree*; the fixes are fencing and
+   completion, not Paxos/Raft.
+
+## 4. Proposed design
+
+### 4.1 Dead-node task completion (Gap A — fixes the live CI hang, issue #896)
+
+Ownership of "complete the origin with an error" must live in exactly one
+place: the send map.
+
+- `SetDead(node)` walks `send_map_` and completes every in-flight task whose
+  target is the dead node with the #628 network-timeout RC, via the same
+  `ClaimOrigin()` linearization used by the exactly-once completion path
+  (PR #878), so a late response cannot double-complete.
+- `SendIn` to a node already marked dead completes the task immediately with
+  the same RC instead of parking it in a retry queue ("skip" today = park
+  forever).
+- Broadcast/replicated tasks: a dead replica contributes its timeout RC to
+  aggregation instead of blocking it; the origin sees a partial-failure RC and
+  can retry against the post-recovery layout.
+- `ScanSendMapTimeouts` becomes the backstop (bounded deadline), not the
+  primary mechanism, and it must *complete* the origin, not only log.
+
+Acceptance: kill a node, submit a task routed to it → `Wait()` returns nonzero
+within the retry deadline. This alone should un-hang the leader_elect test.
+
+### 4.2 Membership epochs + fencing (Gap B)
+
+Introduce a monotonically increasing **membership epoch** (u64), bumped by the
+leader on every confirmed membership change (death or join), carried in the
+host table gossip and stamped on:
+
+- `TriggerRecovery` / container-remap broadcasts: receivers reject stamps
+  older than their current epoch (`kStaleEpochRc`). A stale leader's recovery
+  is refused everywhere rather than half-applied.
+- Recovery idempotence: `(epoch, dead_node_id)` identifies one recovery
+  round; a re-elected or duplicate leader re-running the same round is a
+  no-op, two leaders in different epochs cannot interleave destructively.
+- `UpdateContainerNodeMapping`: mappings record the epoch that produced them;
+  an older-epoch update never overwrites a newer one (today: last-writer-wins
+  by arrival order).
+
+This is fencing, not consensus: nodes may still disagree transiently, but a
+loser's actions are refused deterministically instead of corrupting state.
+
+### 4.3 Failure-detection hardening (Gap C)
+
+- **Priority heartbeats:** SWIM probe/ack handling moves to (or is mirrored
+  on) the dedicated network worker so a starved task-worker pool cannot delay
+  acks. A node that can move bytes is not dead.
+- **Cheap liveness cross-check before `SetDead`:** the confirming node
+  verifies the TCP connection is actually refused/unreachable (transport-level
+  probe) rather than relying purely on task-level ack timeouts. Process death
+  (the real signal in leader_elect: `shm_attach` ENOENT, connection refused)
+  is distinguishable from slowness in one syscall.
+- Timeouts stay configurable per deployment (`swim:` YAML block, already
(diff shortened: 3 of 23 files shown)


lukemartinlogan:
### Update: recovery-path corruption characterized (still open)

Pushed `a4d90a20` — recovery claiming is now **process-wide** instead of per-container. Two reasons beyond the crash:

- A node can host **more than one admin container** (after a prior recovery the survivor owns its own plus the dead node's), so per-instance dedup let each instance recover the same dead node twice.
- Recovery re-creates and re-registers admin containers **while `HeartbeatProbe` fibers are running on them**, making per-instance members precisely the memory pulled out from under the probe.

**Diagnostic payoff.** The previous member-`std::mutex` build aborted with

```
Fatal glibc error: tpp.c:86 (__pthread_tpp_change_priority): assertion failed
```

inside `std::mutex::unlock()`. A default `std::mutex` only reaches glibc's priority-protection path **if its object memory is corrupt** — direct proof that the admin `Runtime` instance's tail region is being clobbered. Moving the state out removed that symptom and the corruption resurfaced as `free(): double free detected in tcache 2`.

**The victim moves with build layout** (invalid pointer → tpp assertion → double free) = wild write / use-after-free near the container instance, not a specific bad free.

Ruled out with instrumentation: dangling `this` (canaries intact), exception unwinding across the fiber boundary (try/catch caught nothing), cross-thread fiber resume (`MIGRATED=0`).

Next: valgrind/ASan on a single-node recovery repro. Leading suspect is `ContainerHold` being a raw pointer into ModuleManager-owned storage while `RegisterContainer` assigns over `info.containers_[container_id]`.

Also hardened the verification harness after it fooled me twice: it now aborts if the libraries under test are older than the newest source edit, and counts **any** glibc abort rather than a single spelling (the old counter reported `crashes=0` while a double-free was happening).

lukemartinlogan:
## leader_elect Phase 1 is GREEN (3/3, zero crashes). Phase 2 remains broken (pre-existing).

### Phase 1 — fixed

| | Result |
|---|---|
| Before this branch | never passed; nodes exited 139 with heap corruption; CI step hit its 15-minute timeout |
| After the corruption + placement fixes | pass=1 fail=2 (hangs), crashes=0 |
| **Now** | **pass=3 fail=0, crashes=0** |

Verified with the suite's own runner (`run_tests.sh`), not just my loop: `[SUCCESS] Phase 1: Leader failover test passed`. Regression subset 14/14 throughout.

Six bugs behind that one red step:

1. **`recovery_initiated_` data race** — `HeartbeatProbe` runs on different worker threads across periods; two `TriggerRecovery` calls mutated a plain `unordered_set` concurrently and the rehash (bucket_count **1 → 13**) freed the bucket array under the other thread.
2. **Dead-node tasks hung forever** (#896) — the dead-node sweep only examined `Physical`-mode queries, but `Dynamic`/`Broadcast` resolve to `DirectId`/`Range`, so nobody completed those origins.
3. **`AggregateOut` delegating to `Copy()`** in 4 broadcast-reachable tasks — overwrote the origin's `task_id_`/`pool_query_`/`completer_` and re-assigned shm strings across segments. ~73 latent instances → #915.
4. **Admin pool redistributed during recovery** — per-node control-plane containers moved onto survivors that already had one.
5. **Pools placed on dead nodes** — `CreatePool` sized to `all_hosts.size()` with identity container→node mapping, so a create after a node death always assigned a container to the dead node.
6. **A dead node's undeliverable replica failed an otherwise-successful create** — the suite's actual assertion. Forgiveness has to live in `PostWait`, since `HandleTaskProgressResult` applies the dead-replica verdict to the origin *after* the survivors aggregate.

Plus **ASan fiber annotations**: the sanitizer previously died with `DEADLYSIGNAL / nested bug in the same thread` on the first fiber fault, so it could not diagnose anything in a runtime where every task body runs on a fiber. It now runs the whole suite. Ironically it found no memory error — what it did was remove the crash noise and make bug #5/#6 legible.

### Phase 2 — still broken, and NOT a regression

`[leader_elect][post_restart]` restarts node 1's runtime and has a fresh client verify cluster health. It hangs at pool creation on the restarted node (900s cap hit).

It was already failing before this branch — then it could not even initialize a client (`REQUIRE(success)` at line 186). It now initializes fine and gets as far as creating a pool, so this branch moved it forward without finishing it. Node **rejoin** is a distinct path from node **death**, and nothing here targeted it.

### Caveat on the last commit

Relaxing the suite's SWIM timings (2s/2s/3s → 5s/5s/8s) is a **test-config** change. It is justified — under docker load those values declared *healthy* nodes dead (replicas "Gone" on two live nodes with zero crashes), after which the survivors self-fenced and the create never completed — and the constraint that forced them is gone now that creates tolerate an undeliverable replica. But it does not fix the underlying fragility: SWIM still reads CPU starvation as death, and self-fencing still disables recovery outright on a small cluster. That is Phase 4 of `docs/leader_election_failover.md`, not something this branch solves.

### Still to do before this leaves draft

- [ ] Phase 2 (post-restart / node rejoin)
- [ ] Full `ctest` suite (only a 14-test subset has been run)
- [ ] CI on the branch

lukemartinlogan:
### Full local test sweep: 293/302 pass, zero code failures

Ran the complete `ctest` suite (not just the CTE subset I had been using), since this branch touches fiber machinery, pool geometry, dead-node completion, and four task types' `AggregateOut`.

- **293 of 302 pass.**
- **All 9 failures are `*_integration_docker` wrappers**, and they fail for a local environment reason, not a code one:

```
mount denied: the source path "\\Users\\llogan\\...:c:\\Users\\llogan\\..." too many colons
```

A bare `ctest` does not set `HOST_WORKSPACE`, so the compose files fall back to a Windows-style path Docker cannot mount on this WSL box. Run with `HOST_WORKSPACE` set, those same suites work — that is how `leader_elect` Phase 1 was taken to 3/3.

- **Zero non-docker failures.**

Also reverted in `5b4bdc33`: the live-node pool sizing I had added. `CreatePool` runs on every node handling the broadcast, so each computed its own container count from its own transient liveness view — Phase 2 caught the restarted node building the pool with 4 containers while the survivors built 3, a split-brain layout. Pool geometry is deterministic again (hostfile size, identity container→node map), with comments at both sites explaining why liveness must not be consulted there. Phase 1 still passes 3/3 afterwards, so the green result rests on the dead-replica forgiveness and the SWIM timings, not on placement.

### Remaining before this leaves draft

- [ ] **Phase 2 / node rejoin** (#31 in my notes): `[leader_elect][post_restart]` hangs creating a pool on the restarted node. Pre-existing — before this branch it failed even earlier, unable to initialize a client at all. Node rejoin is a different path from node death and nothing here targeted it.
- [ ] **CI on the branch.**

lukemartinlogan:
## ✅ Cluster Tests GREEN in CI — including the leader-election step

Run [31005222978](https://github.com/iowarp/clio-core/actions/runs/31005222978): **`completed/success`**

```
completed/success   Runtime leader-election test (leader death + re-election)
completed/success   CTE cache coherence test (4-node, issue #886)
```

This is the step that was hitting a 15-minute timeout on dev and blocking the dev→main PR. It now passes in CI's own environment, not just locally (where it runs 3/3 with zero crashes).

**Note for future branches:** `cluster-tests.yml` triggers only on `push` to `main`/`dev` plus `workflow_dispatch` — it does **not** run on branch PRs. A branch can therefore look entirely green while the suite it fixes was never executed. I dispatched it manually against this branch; anyone touching cluster behaviour should do the same rather than trust the PR check set.

### Remaining PR checks

- **`build-test (windows-2025)`** — `cte_bdev_leak_stress` **Timeout**, 226/227 Windows tests pass. Known flake: it fails on dev too, and there is prior commit history settling this same test. Unrelated to anything here (leader election, recovery, fibers, pool geometry).
- **`codecov/patch`** — coverage on new lines. Most of this branch is distributed recovery code that only executes in multi-node docker clusters, which do not feed coverage. I added a single-node regression guard for the `AggregateOut` contract (origin identity survives aggregation; `num_recovered_` sums rather than being clobbered) — both assertions fail against the old implementation — but it does not clear the threshold on its own.

Everything else: **36 pass**.

lukemartinlogan:
## Both leader_elect phases now pass — including Phase 2, which had never passed

```
BOTH-PHASE RESULT pass=2 fail=0
iter1 rc=0 phase1_pass=1 phase2_pass=1 aborts=0
iter2 rc=0 phase1_pass=1 phase2_pass=1 aborts=0
```

Cluster Tests dispatched on this branch: **completed/success** (leader-election *and* coherence). Full local `ctest`: **297/302 with zero code failures** — the 5 remaining are docker wrappers looking for `/workspace/build` while this box builds into `build-cpu`; CI builds into `build`, so they pass there.

### Phase 2 root cause: death was permanent

SWIM's probe loop skipped dead nodes outright, so a restarted node could only be noticed by a peer that happened to receive traffic from it (`RecvIn`'s `SetAlive`). Any peer without direct traffic kept a stale `kDead` forever and silently dropped that node's responses. The post-restart client runs **on** the rejoined node, so a peer still believing it dead never routed the reply back — the hang. Observed directly: node2 held node0 `DEAD` while node3/node4 had it `ALIVE`.

Three parts, all required:
1. **Dead nodes are probed**, so rejoin is observable at all.
2. **A successful probe revives via `SetAlive` + `FlushStaleStateForNode`**, not bare `SetNodeState` — the latter only flips the enum and leaves the node in `dead_nodes_`, which the dead-node completion sweep consults, so a "revived" node kept having its tasks failed.
3. **A failed probe must never downgrade a dead node.** Probing dead nodes initially moved them to `kProbeFailed`/`kSuspected`, taking them *out* of `kDead` and disabling the completion sweep — which re-broke Phase 1. The two-phase gate caught it.

### Starvation is no longer read as death (the real Phase 4 fix)

`Host::last_inbound` records when any traffic last arrived from a peer, and the suspicion→dead promotion is vetoed while that traffic is recent, with a log line saying so. Bytes from a peer prove it is up regardless of what the probe did. This replaces the test-timing mitigation with an actual product fix.

### #915: contract enforced, and a self-inflicted regression caught

All 7 module-template `AggregateOut` implementations fixed — that file is the propagation vector for every new module.

⚠️ **My first attempt at this broke streaming**, and CI's `leak-check` caught it (`cr_streaming_force_net`: the 1MB output came back empty). I had written a comment claiming those template tasks "have no OUT fields" without verifying — **5 of 7 do** (`data_`, `current_depth_`, `result_value_`, `counter_value_`), so dropping `Copy()` silently discarded exactly the data they exist to return. Each now merges its real fields; streaming suite 5/5.

That is the concrete argument for **not** blind-editing the ~26 remaining instances in bdev/compressor/replication/cache/CAE: bdev's `AggregateOut` genuinely merges `blocks_` through `Copy`, and the same mistake there corrupts I/O rather than failing a test. They are instead audited as single-replica-routed (`DirectHash` resolves to one container) **and guarded at runtime**: `RecvOutAggregate` snapshots the origin's `task_id_` across the call and, if a task mutates it, logs a violation naming pool and method and restores the identity.

lukemartinlogan:
## All four design-doc phases now implemented and verified

```
BOTH-PHASE RESULT pass=2 fail=0
iter1 rc=0 phase1_pass=1 phase2_pass=1 aborts=0
iter2 rc=0 phase1_pass=1 phase2_pass=1 aborts=0
regression subset (incl. streaming/force_net + autogen): 24/24
```

| Phase (docs/leader_election_failover.md) | Status |
|---|---|
| 0 — recovery crash (#856) | fixed |
| 1 — dead-node task completion (#896) | fixed |
| 2 — client failover polish | **fixed** |
| 3 — membership epoch + fenced recovery | **fixed** |
| 4 — failure-detection hardening | fixed |

### Phase 2 and Phase 3 each fixed a real bug, not just polish

**Client failover sampled hosts uniformly at random** from the whole hostfile, so it could burn every one of its 16 attempts re-dialling the node that had just died — and post-failover behaviour was irreproducible under test. Order is now deterministic: presumptive leader first (it coordinates recovery), then alive hosts, then believed-dead ones last as a fallback rather than excluded, since "dead" is a local opinion that goes stale the moment a peer rejoins.

**Recovery claims were keyed by dead-node id alone**, which is wrong in both directions:
- A node that died, **rejoined, and died again could never be recovered a second time** — its id was already claimed, forever. The rejoin fix earlier in this branch makes that sequence reachable, so this became live rather than theoretical.
- "Leader" is not consensus — every node computes `lowest alive id` from its own SWIM view — so during churn two nodes can briefly both believe they lead, with nothing to fence the loser.

`IpcManager` now carries a monotonic membership epoch, bumped on every confirmed `SetDead`/`SetAlive`, and claims are keyed by `(epoch, dead node)`: a later death is a distinct recoverable event, while a duplicate or stale coordinator inside one epoch is still refused exactly once.

Issue #929, opened 2026-08-06
dev→main PR flakes ~47% of runs: four independent causes, one of them a duplicated PoolManager singleton (root cause of #923 + #928)


lukemartinlogan:
## Summary

The `dev` → `main` PR (#908) is red more often than green, but **not for one reason**. There are four independent flake classes. One of them — a genuine data race that produces **two `PoolManager` instances in a single process** — is the root cause of three separately-filed issues (#923, #928, and the current `windows-11-arm` failure).

Measured over the last 200 workflow runs on `dev`:

| Workflow | runs | failures | rate |
|---|---|---|---|
| Cluster Tests | 12 | 4 | **33%** |
| CI macOS | 25 | 2 | 8% |
| CI Linux | 25 | 2 | 8% |
| CI Windows | 26 | 2 | 7% |
| CI Adapters / CI VFD / GPU / cpplint / Build Core Containers | 26/26/21/26/13 | 0 | 0% |

A dev→main merge needs all of these green *simultaneously*. At the rates above that is `0.67 × 0.92 × 0.92 × 0.93 ≈ **53%**` — so roughly **one in two attempts goes red on its own**, with no code change required. That is the "flaking".

---

## Class A — duplicated `PoolManager` singleton (the big one)

This is the same defect as **#923** (Linux `leak-check`, `cte_reorganize_force_net`) and **#928** (macOS `fuse_cte_cross_page`), and it just took down `build-test (windows-11-arm)` on PR #908.

Symptom is always identical — the admin pool (`major:1`) is never routable, and the #849 inline-retry deadline burns its full 30 s before failing the task with `(u32)-1` = `4294967295`:

```
RouteLocal: Container not found for pool=PoolId(major:1, minor:0) container_id=0 method=10
            (pool_manager=..., initialized=0, pools known=0)
RouteTask: inline retry exhausted after 30007 ms (budget 30000 ms) on non-worker thread
```

### The diagnostic added in #923 identifies it definitively

`ipc_manager.cc` deliberately logs the `PoolManager` **instance address** at the miss site, and `PoolManager::ServerInit` logs the address it initialized, precisely so the two causes can be told apart. Per that comment: *"two addresses is duplication, one address is ordering."*

**windows-11-arm** — [job 92499029730](https://github.com/iowarp/clio-core/actions/runs/31064399505/job/92499029730), `cr_safe_bdev_concurrency_tests`:

```
PoolManager::ServerInit: instance=00000261591FA710      <- initialized here
[...]

PR #938, opened 2026-08-07
Build the HDF5 VFD off the UNIX gate, not the ELF gate

### LINKED ISSUE #937: CTE VFD is gated behind CLIO_CORE_ENABLE_ELF, so it is never built on macOS
## Summary

`add_subdirectory(vfd)` sits inside `if(CLIO_CORE_ENABLE_ELF)` in
`context-transfer-engine/adapter/CMakeLists.txt`, and `CLIO_CORE_ENABLE_ELF`
does `pkg_check_modules(libelf REQUIRED libelf)` (CMakeLists.txt:1394). That
cannot be satisfied on Mach-O, so the `clio_vfd` target is never created on
macOS and `cmake --build --target clio_vfd` fails with
`No rule to make target 'clio_vfd'`.

The gate looks incidental rather than deliberate:

* The VFD is a plugin HDF5 `dlopen`s, not a libc interceptor. `H5FDclio.cc`
  never includes `real_api.h` and never calls `dlsym`/`RTLD_NEXT`.
* It links `ctp::interceptor`, but that is an INTERFACE target which only
  attaches `${ELF_LIBS}` when `CLIO_CORE_ENABLE_ELF` is on
  (`context-transport-primitives/CMakeLists.txt:164-174`); otherwise it just
  defines `CTP_ENABLE_ELF=0`.
* Since e2050b6b ("Delete adapter/cfs; the filesystem client owns the
  descriptor layer") its remaining dependency is `clio_cte_filesystem_client`,
  which is not gated on ELF or UNIX at all.

The same file already draws this distinction for `cfs`, in the comment above
its gate: *"Gated on UNIX, not on ELF: unlike the interceptors it never touches
real_api.h."* That reasoning applies to the VFD as well.

## Evidence

I moved the VFD from the ELF gate to a UNIX gate in CI (nothing else changed)
and ran it on `macos-26`:

1. **It builds.** `H5FDclio.cc` compiles and links with
   `CLIO_CORE_ENABLE_ELF=OFF`, producing `libclio_vfd.dylib`:
   https://github.com/hyoklee/hpf/actions/runs/31141294898

2. **It works.** Against HDF5 `develop` (built from source) and netCDF-C
   `main`, with a live `clio_run`, `tst_chunks3` driven through
   `HDF5_DRIVER=clio_vfd` produced all 18 measurements:
   https://github.com/hyoklee/hpf/actions/runs/31142087871

   `otool`/`LC_RPATH` was checked to confirm the plugin resolved the
   benchmark's own libhdf5 rather than the conda one, so this is a real VFD
   run and not a silent fallback to sec2:

   ```
   libclio_vfd.dylib resolves libhdf5.1000.dylib -> .../nc4-clio-work/hdf5-install/lib/libhdf5.1000.dylib
   clio_run ready
   ==> Running variant: clio_vfd   → ok
   ```

## Proposed change

```cmake
if(UNIX AND CLIO_CTE_ENABLE_VFD)
    add_subdirectory(vfd)
endif()
```

next to the other non-interceptor pieces, and drop the `CLIO_CTE_ENABLE_VFD`
block from the ELF section. Windows stays excluded, matching
`ci-adapters.yml`, which already sets `CLIO_CTE_ENABLE_VFD=OFF` there —
`H5FDclio.cc` uses `<sys/file.h>`, `<unistd.h>` and `flock`, so Windows needs
real porting work rather than a gate change.

## Two unrelated things noticed on the way

* e2050b6b left a dead block in the same file:

  ```cmake
  if(UNIX)
      endif()
  ```

  Valid CMake, does nothing — the `add_subdirectory(cfs)` line it wrapped is
  gone. Happy to remove it in the same PR or leave it alone.

* On macOS the VFD shows the same contiguous-write cost as on Linux, about 7x
  more pronounced: `contiguous write 64 64 1` took **470 s** against 0.010 s
  for the native baseline, while every chunked and compressed case stayed
  within 1.5-4x of native. It scales with the number of `nc_put_vara` calls
  rather than bytes moved, which suggests a per-request round trip. Not
  addressed here — say the word and I will open a separate issue with the full
  table.

hyoklee:
Closes #937.

`add_subdirectory(vfd)` sat inside `if(CLIO_CORE_ENABLE_ELF)`, and that option
does `pkg_check_modules(libelf REQUIRED libelf)` — unsatisfiable on Mach-O — so
the `clio_vfd` target was never created on macOS.

Nothing in the VFD needs ELF:

* It is a plugin HDF5 `dlopen`s, not a libc interceptor. `H5FDclio.cc` never
  includes `real_api.h`, never calls `dlsym`/`RTLD_NEXT`.
* It links `ctp::interceptor`, but that INTERFACE target only attaches
  `${ELF_LIBS}` when ELF is on; otherwise it just defines `CTP_ENABLE_ELF=0`.
* Since e2050b6b its remaining dependency is `clio_cte_filesystem_client`,
  which is not gated at all.

This is the distinction the file already made for `cfs` before that commit
removed it: *"Gated on UNIX, not on ELF: unlike the interceptors it never
touches real_api.h."*

### Verification

Both runs applied exactly this change to `dev` and nothing else, on `macos-26`:

| | result |
| --- | --- |
| [builds](https://github.com/hyoklee/hpf/actions/runs/31141294898) | `H5FDclio.cc` compiles and links with `CLIO_CORE_ENABLE_ELF=OFF` → `libclio_vfd.dylib` |
| [works](https://github.com/hyoklee/hpf/actions/runs/31142087871) | `tst_chunks3` through `HDF5_DRIVER=clio_vfd`, against HDF5 `develop` + netCDF-C `main` + a live `clio_run`, produced all 18 measurements |

The second run checks `otool`/`LC_RPATH` to confirm the plugin resolved the
intended libhdf5, so it is a real VFD run rather than a silent fallback to
sec2:

```
libclio_vfd.dylib resolves libhdf5.1000.dylib -> .../nc4-clio-work/hdf5-install/lib/libhdf5.1000.dylib
clio_run ready
==> Running variant: clio_vfd   → ok
```

### Scope

Windows stays excluded through the `UNIX` gate, matching `ci-adapters.yml`,
which already sets `CLIO_CTE_ENABLE_VFD=OFF` there — `H5FDclio.cc` uses
`<sys/file.h>`, `<unistd.h>` and `flock`, so it needs porting work rather than
a gate change.

The new block also replaces the empty `if(UNIX) endif()` that e2050b6b left
behind when `add_subdirectory(cfs)` was removed. Happy to split that out if you
would rather keep this to one thing.

Not addressed here, but noted in #937: with the VFD running on macOS,
`contiguous write 64 64 1` takes ~470 s against 0.010 s for the native
baseline, while chunked and compressed access stays within 1.5-4x. It tracks
the number of `nc_put_vara` calls rather than bytes moved. I can open a
separate issue with the full table if that is useful.

### DIFF
--- a/context-transfer-engine/adapter/CMakeLists.txt
+++ b/context-transfer-engine/adapter/CMakeLists.txt
@@ -5,8 +5,13 @@
 # CTE adapter singleton implementation - moved to filesystem directory
 # to leverage existing working build configuration
 
-if(UNIX)
-    endif()
+# The HDF5 VFD is a plugin the library dlopen's, not an interceptor: it
+# never touches real_api.h, so it does not need ELF support. Gate it on
+# UNIX -- H5FDclio.cc uses <sys/file.h>, <unistd.h> and flock, so Windows
+# needs porting work rather than a gate change.
+if(UNIX AND CLIO_CTE_ENABLE_VFD)
+    add_subdirectory(vfd)
+endif()
 
 # Adapters require ELF support (for real_api.h)
 # Disable all adapters when ELF is disabled
@@ -22,9 +27,6 @@ if(CLIO_CORE_ENABLE_ELF)
     endif()
     if (CLIO_CTE_ENABLE_MPIIO_ADAPTER)
         add_subdirectory(mpiio)
-    endif()
-    if (CLIO_CTE_ENABLE_VFD)
-        add_subdirectory(vfd)
     endif()
     if (CLIO_CTE_ENABLE_NVIDIA_GDS_ADAPTER)
         add_subdirectory(nvidia_gds)
--- a/context-transfer-engine/filesystem/CMakeLists.txt
+++ b/context-transfer-engine/filesystem/CMakeLists.txt
@@ -26,8 +26,12 @@ target_include_directories(clio_cte_filesystem_runtime PRIVATE
 target_include_directories(clio_cte_filesystem_client PRIVATE
     ${CMAKE_SOURCE_DIR}/context-transfer-engine/core/include)
 
-# Export dllexport for the g_fs_client global on Windows (reuses clio_cte/api.h).
+# Export the g_fs_client global on Windows. This must be THIS module's macro:
+# defining CLIO_CTE_CORE_BUILDING_DLL here also told the core's headers that
+# g_cte_client was being defined locally, which it is not -- an LNK2019 that
+# stayed hidden while the descriptor layer (its only user here) was excluded
+# from Windows builds.
 if(TARGET clio_cte_filesystem_client)
   target_compile_definitions(clio_cte_filesystem_client PRIVATE
-    CLIO_CTE_CORE_BUILDING_DLL=1)
+    CLIO_CTE_FS_BUILDING_DLL=1)
 endif()
--- a/context-transfer-engine/filesystem/include/clio_cte/filesystem/api.h
+++ b/context-transfer-engine/filesystem/include/clio_cte/filesystem/api.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2024, Gnosis Research Center, Illinois Institute of Technology
+ * All rights reserved. BSD 3-Clause license.
+ */
+#ifndef CLIO_CTE_FILESYSTEM_API_H_
+#define CLIO_CTE_FILESYSTEM_API_H_
+
+#include "clio_cte/api.h"
+
+/**
+ * Per-DLL export macro for clio_cte_filesystem_client.
+ *
+ * Mirrors clio_cte/api.h's CLIO_CTE_API and clio_runtime/api.h's CLIO_RUN_API.
+ * CMake sets CLIO_CTE_FS_BUILDING_DLL=1 PRIVATE on the
+ * clio_cte_filesystem_client target; consumers see the import form.
+ *
+ * This exists rather than reusing CLIO_CTE_API because that macro answers a
+ * different question. CLIO_CTE_API means "is this translation unit building
+ * clio_cte_core_client?", so defining CLIO_CTE_CORE_BUILDING_DLL here to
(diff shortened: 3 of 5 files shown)


hyoklee:
The two red checks here are pre-existing on `dev` and are not caused by this
change. Both reproduce on re-run, so they are deterministic rather than flaky:

| check | failing test | same failure on `dev`? |
| --- | --- | --- |
| `leak-check` (CI Linux) | `cte_client_config_force_net` | yes — [run 31124646795](https://github.com/iowarp/clio-core/actions/runs/31124646795), same test |
| `icx (windows-2025)` | `cte_reorganize_all` | yes — the latest `dev` CI Windows run fails `cte_reorganize_all` too |

The leak-check failure is a genuine runtime-heap leak, not a harness problem:

```
[FAIL] Client - AsyncGetTargetInfo [core][client][target]
  Runtime-heap leak in test 'Client - AsyncGetTargetInfo': 147624 bytes still
  allocated after the test (before=884246, after=1031870)
```

Neither test can be reached by this diff, which only moves
`add_subdirectory(vfd)` between two CMake conditionals —
`cte_client_config_force_net` is a runtime config test and `cte_reorganize_all`
is a CTE reorganisation test, and neither loads the VFD. (`cte_reorganize_all`
is already in the macOS CI's `-E` exclusion list.)

Everything that this change could plausibly affect is green: `vfd (linux,
native write-through)`, `adapters (linux, all 5)`, `build-test (macos-15)`,
`build-test (windows-2025)`, `build-test (unit amd64)`, `sanitizer (asan)` and
`cpplint`.

Related to #929, though note that issue measures these as intermittent (CI
Linux 8%, CI Windows 7%) while both now reproduce every run — the
`AsyncGetTargetInfo` leak in particular looks like a real regression worth its
own issue rather than a flake. Happy to open one and dig into it separately if
that would help.

PR #941, opened 2026-08-07
Consolidate gpu_vector branches; fix the in-kernel fault deadlock

lukemartinlogan:
## Summary

Consolidates the two diverged gpu_vector branches into one, and carries this
round's fault-path work. **Read the "What is broken" section before merging —
5 tests fail.**

`gpu-vector-weight-resume` and `gpu-vector-weight-resume-v2` had diverged at
`c95cecdb`: 109 commits existed only on the first, 90 only on the second.
Neither was a superset, so work was silently present or absent depending on
which branch was checked out. This branch is the union.

## What is broken

Five gpu_vector suites, all from the OLD branch, fail against the newer vector:

| test | status |
|---|---|
| `cte_gpu_vector_tensor_map_cuda` | Failed (2.6s) |
| `cte_gpu_vector_compress_fault_cuda` | Failed (3.6s) |
| `cte_gpu_vector_huge_compress_cuda` | Failed (285s, likely a hang) |
| `cte_gpu_vector_transaction_cuda` | Failed (3.7s) |
| `cte_gpu_vector_prefetch_cuda` | Failed (3.6s) |

These are the old branch's tests meeting v2's redesigned fault path (per-lane
page pointers, block-collective faults, in-kernel HBM resolve). They were
written against the older warp-collective contract. I have NOT yet determined,
per test, whether the test encodes an obsolete contract or the new vector
genuinely regressed a behaviour the old one had. That triage is the next step
and should happen before this merges.

Separately, a wider sweep shows 6 CTE failures (`cte_functional_all`,
`cte_tag_reorganize`, `cte_interposer_stack`, `cte_cache_stack`,
`cte_data_organizer_all`, `cte_functional_force_net`) plus 2 compressor ones.
The 2 compressor failures are confirmed pre-existing. The 6 CTE ones are
**not caused by this branch** — verified by controlled A/B, identical failures
with chunked compression on/off and with the bdev prefault on/off — but I could
not build the merge-base to prove they are pre-existing (two attempts died on
environment issues: CUDA detection in the shared build dir, then a SYCL/zfp
link mismatch in a fresh one).

## What works

- Builds clean: 0 errors across the whole tree.
- 13 of 18 in the targeted run pass, including all 5 CAE/compressor/pagify
  suites and `cte_gpu_vector_{cuda,matrix,oversubscribe,llama_geom,gguf_map,
  compress,capacity}`.
- Benchmark `clio_compressed_fault_bench` passes in every mode with byte-exact
  checksums: uncompressed in-kernel 11–12 GB/s at 32/64/128 blocks, compressed
  in-kernel 6.0–6.8 GB/s, host-orchestrated mode 1 at 7.5 GB/s.

## The substantive work in here

**The in-kernel fault deadlock is fixed.** The one-launch device-vector path
hung nondeterministically. Root cause, isolated by controlled comparison rather
than inspection: the hang requires the fault to be served from the DEVICE tier.
Same workload, same block count — a 32 MB device tier (faults served from host
over the DMA copy engine) ran 3/3 clean; a 700 MB one (served device-to-device)
hung 3 of 4. A D2D copy needs SM time, the persistent kernel spins on every SM
waiting for exactly that copy, and neither side can proceed. The fix removes the
host from that path: `BuildDevicePageTable` publishes a directory of pages
already in VRAM and the faulting block copies them itself — no task, no runtime
round trip. Four hypotheses were tested and disproved along the way (lazy
`cudaMalloc` on the fault path, worker starvation, an unbounded evict loop,
default-flag streams); the first was a real latent deadlock and is fixed too.

**Compressed pages decode in-kernel**, taking that path from hanging → 1.3 GB/s
→ 6.8 GB/s. The last 5x came from the stored form, not the decoder: host lz4
wrote one LZ4 block per page, and LZ4 decoding is inherently serial, so one
thread decoded 4 MiB while 255 idled. Pages are now stored as independent 64 KiB
chunks so one thread per chunk decodes in parallel. The chunked form uses its
own magic (`HB4C`) — reusing `raw_lz4`'s `RB4C` broke two tests immediately
because several readers detect an nvcomp frame by that magic alone.

**pagify** gives the CAE a unified page-number API: `PageMap` (arithmetic) plus
`Pagify` (fetch and construct), served through the CAE's existing CTE
interceptor, so a `GetBlob` named `page:<id>` is resolved from the tag's
`pagemap` blob and built from blobs the caller never names. Handles mixed pages
(several blobs), partial pages, and gaps that must read as zero. Opt-in per
request — anything not named `page:<id>` forwards unchanged.

## Test plan

- [x] Full tree builds, 0 errors
- [x] Benchmark passes all modes, byte-exact checksums
- [x] CAE/pagify/compressor-interpose suites pass
- [ ] **Triage the 5 failing old-branch gpu_vector suites**
- [ ] Establish a clean merge-base build to close out the 6 CTE failures

## Breaking changes

Host-compressed payloads are now stored chunked (new `HB4C` magic). Readers
detect the form by magic and both are supported, so existing data still reads;
but data written by this branch is not readable by an older build.


lukemartinlogan:
## CORRECTION to the attribution in this PR's description

The original description claimed the 6 CTE failures were "not caused by this
branch" and that the 2 compressor failures were "confirmed pre-existing".
**Both claims were wrong.** I have now built a branch that is `dev` plus
nothing but the deletion of the gpu_vector adapter, and run the same suites.

Corrected results:

| test | on dev | on this PR | verdict |
|---|---|---|---|
| `cte_functional_all` | PASS | FAIL | **broken by this branch** |
| `cte_tag_reorganize` | PASS | FAIL | **broken by this branch** |
| `cte_interposer_stack` | PASS | FAIL | **broken by this branch** |
| `cte_data_organizer_all` | PASS | FAIL | **broken by this branch** |
| `cte_functional_force_net` | PASS | FAIL | **broken by this branch** |
| `cte_cache_stack` | FAIL | FAIL | genuinely pre-existing |
| `compressor_error_handling` | does not exist | FAIL | added by this branch, failing |
| `compressor_transparent_getblob` | does not exist | FAIL | added by this branch, failing |

Why the earlier attribution was wrong, so the mistake is not repeated:

1. The A/B I ran toggled only TWO changes (chunked compression, the bdev tier
   prefault). Both were cleanly exonerated -- the failures are identical with
   each on and off -- but I then generalised that to "not caused by this
   branch", which the experiment does not support. Something else in the branch
   is responsible.
2. The `git stash` test that "confirmed" the compressor failures pre-existing
   removed only UNCOMMITTED edits. Everything else on the branch was already
   committed, so the stashed build still contained it. The test proved nothing
   about the branch.

Net: this PR breaks 5 CTE tests that pass on dev, and adds 2 failing tests, on
top of the 5 failing gpu_vector suites already listed. The cause of the 5 CTE
regressions is not yet identified; the A/B only rules out chunked compression
and the tier prefault.

Given this, the branch should not merge as-is.

Issue #970, opened 2026-08-13
HDF5 VOL: process never exits — clio's atexit teardown runs before H5_term_library, then clio_file_close waits forever on a dead client


hyoklee:
## Summary

With `HDF5_VOL_CONNECTOR=clio`, any application that still holds an HDF5 file open when `main` returns **never exits**. All I/O has already completed and the runtime is healthy; the process just sits in `exit()` forever. There is no timeout anywhere on the path, so nothing recovers it — only an external kill.

Every netCDF-4 application hits this, because HDF5 is left holding the file at exit even when the application called `nc_close`.

## Reproducer

No netCDF needed — 30 lines, and the only thing that matters is *not* calling `H5Fclose`:

```c
/* vol_exit_hang.c */
#include <hdf5.h>
#include <stdio.h>

int main(void) {
    hsize_t dims[1] = {16};
    int buf[16];
    hid_t f, s, d;

    for (int i = 0; i < 16; i++) buf[i] = i;

    f = H5Fcreate("vol_exit_hang.h5", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
    if (f < 0) { printf("H5Fcreate failed\n"); return 1; }
    s = H5Screate_simple(1, dims, NULL);
    d = H5Dcreate2(f, "d", H5T_NATIVE_INT, s, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
    H5Dwrite(d, H5T_NATIVE_INT, H5S_ALL, H5S_ALL, H5P_DEFAULT, buf);
    H5Dclose(d);
    H5Sclose(s);
    /* no H5Fclose(f) -- H5_term_library will close it at exit */

    printf("returning from main\n");
    fflush(stdout);
    return 0;
}
```

```console
$ cc -o vol_exit_hang vol_exit_hang.c -lhdf5
$ clio_run start &                     # cte_core + bdev + cte_filesystem composed
$ export HDF5_PLUGIN_PATH=<clio>/bin HDF5_VOL_CONNECTOR=clio
$ timeout 90 ./vol_exit_hang
...
returning from main
... ipc_manager.cc:2396 INFO ClearClientPool [ClientPool] Clearing 0 persistent connections
$ echo $?
124                                    # killed by timeout, 90s later

# same program with H5Fclose(f) added:
exit=0  elapsed=1s
```

`ClearClientPool` is the last thing printed, which is the clue: that line is inside `IpcManager::ClientFinalize`.

[...]

PR #971, opened 2026-08-13
fix(hdf5): make the VOL's coherence stamp portable off glibc (macOS + Windows build failure)

hyoklee:
## What is broken

`clio_hdf5_vol` has not compiled on macOS or Windows since `6873b60a` added the
coherence stamp. The stamp reads `st.st_mtim` and calls
`clock_gettime(CLOCK_REALTIME, ...)`, and neither exists on both other
platforms:

```
macOS (clang, macos-26):
  clio_vol.cc:822:51: error: no member named 'st_mtim' in 'stat'
  clio_vol.cc:823:51: error: no member named 'st_mtim' in 'stat'
  clio_vol.cc:908:52: error: no member named 'st_mtim' in 'stat'
  clio_vol.cc:909:52: error: no member named 'st_mtim' in 'stat'

Windows (MSVC 19.4x, windows-2025):
  clio_vol.cc(822,51): error C2039: 'st_mtim': is not a member of 'stat'
  clio_vol.cc(823,51): error C2039: 'st_mtim': is not a member of 'stat'
  clio_vol.cc(905,21): error C2065: 'CLOCK_REALTIME': undeclared identifier
  clio_vol.cc(905,7):  error C3861: 'clock_gettime': identifier not found
  clio_vol.cc(908,52): error C2039: 'st_mtim': is not a member of 'stat'
  clio_vol.cc(909,52): error C2039: 'st_mtim': is not a member of 'stat'
```

`struct stat` does not agree across the three platforms: POSIX-2008 and glibc
name the `struct timespec` member `st_mtim`, Darwin predates that name and calls
the same member `st_mtimespec`, and MSVC's `struct stat` has no sub-second
member at all — only `st_mtime`, in whole seconds.

## What this changes

- **`clio_stat_mtime(st, &sec, &nsec)`** — one place that knows the three
  spellings; every stamp read goes through it. It hands back the two halves
  separately rather than a single nanosecond count, so the stamp text stays
  exactly `<sec>.<nsec>` and stamps already stored in a tier still compare equal
  on Linux. No cache-layout version bump is needed.
- **`clio_realtime_now_ns()`** using `std::chrono::system_clock` instead of
  `clock_gettime(CLOCK_REALTIME, ...)`. `<chrono>` was already included; the
  clock is the same wall clock on every implementation this builds against, and
  it cannot fail, which retires the "cannot read the clock" arm of
  `clio_stamp_ambiguous`.
- **One behavioural change, Windows only**: the default stamp granularity
  becomes 1 s there. `stat()` on Windows reports mtime in whole seconds (two on
  a FAT-formatted volume), so keeping the 10 ms default would call a file
  unambiguous whose very next write lands in the same reported second — exactly
  the case `clio_stamp_ambiguous` exists to refuse. 1 s is the smallest default
  that still fails closed on that clock. `CLIO_VOL_STAMP_GRANULARITY_NS`
  overrides it as before.

Nothing else in `clio_vol.cc` is platform-specific: `stat()` itself is fine on
MSVC, and those two were the only POSIX-only uses in the file.

## Verification

- **Linux** (glibc 2.39, gcc 13, HDF5 `develop` 2.3.0): `libclio_hdf5_vol.so`
  builds, and `tst_chunks3` from netCDF-C `nc_perf` driven through
  `HDF5_VOL_CONNECTOR=clio` against a live `clio_run` produces all 18 timings,
  identical to before the change. The stamp text is byte-identical on this
  platform — the same `long long` values through the same `std::to_string`.
- **macOS / Windows**: I have no local access to either, so this repo's own
  adapter CI legs are the check for those two branches of the `#if`. The failures
  quoted above are from real runs on `macos-26` and `windows-2025` (hyoklee/hpf
  CI, 2026-08-13), so if this PR's macOS and Windows legs compile
  `clio_hdf5_vol`, the specific breakage is gone.

## How it was found

Measuring the CLIO adapters against netCDF-4 in
[hyoklee/hpf](https://github.com/hyoklee/hpf): the macOS and Windows dashboards
have had **no `clio_vol` series at all since 2026-08-10**, while Linux publishes
one nightly. The macOS VFD series works and is publishing, which is what made the
VOL's absence stand out.

Unrelated and still open, from the same measurements: #970 (the VOL cannot exit
when HDF5 still holds a file at exit) is a separate bug and is untouched here —
the Linux run above still ends in that hang, exactly as it did before.

### DIFF
--- a/context-transfer-engine/adapter/hdf5_vol/clio_vol.cc
+++ b/context-transfer-engine/adapter/hdf5_vol/clio_vol.cc
@@ -30,7 +30,7 @@
 #endif
 
 #include <sys/stat.h>       /* stat() -- coherence stamp */
-#include <time.h>           /* clock_gettime() -- stamp granularity check */
+#include <time.h>           /* time_t, for struct stat's mtime members */
 
 #include <cstdint>
 #include <cstdlib>
@@ -802,6 +802,47 @@ static void clio_resolve_config(hid_t fapl_id, size_t *chunk_size,
    which HDF5 always renders with a leading '/'. */
 static constexpr const char *kStampBlobName = "__clio_coherence_stamp";
 
+/* The file's modification time, split into whole seconds and nanoseconds.
+ *
+ * `struct stat` does not agree across the platforms this connector builds on:
+ * POSIX-2008 (and glibc) names the timespec member st_mtim, Darwin predates
+ * that name and calls the same member st_mtimespec, and MSVC's struct stat has
+ * no sub-second member at all -- only st_mtime, in whole seconds. Reading
+ * st_mtim unconditionally is what stopped this file compiling anywhere but
+ * glibc; every use goes through here instead.
+ *
+ * Callers keep the two halves separate rather than taking a single nanosecond
+ * count, because the stamp string embeds them as "<sec>.<nsec>" and that text
+ * is compared against stamps already stored in a tier. */
+static void clio_stat_mtime(const struct stat &st, long long *sec,
+                            long long *nsec) {
+#if defined(_WIN32)
+  /* No sub-second field exists; see clio_stamp_granularity_ns, which widens the
+     ambiguity window to match this coarser clock. */
+  *sec = static_cast<long long>(st.st_mtime);
+  *nsec = 0;
+#elif defined(__APPLE__)
+  *sec = static_cast<long long>(st.st_mtimespec.tv_sec);
+  *nsec = static_cast<long long>(st.st_mtimespec.tv_nsec);
+#else
+  *sec = static_cast<long long>(st.st_mtim.tv_sec);
+  *nsec = static_cast<long long>(st.st_mtim.tv_nsec);
+#endif
+}
+
+/* Wall-clock now, in nanoseconds since the epoch.
+ *
+ * std::chrono::system_clock rather than clock_gettime(CLOCK_REALTIME): MSVC has
+ * neither the function nor the macro, and system_clock is the same wall clock
+ * on every implementation this builds against. It is also unfailing, which
+ * removes the "cannot read the clock" arm the caller used to need. */
+static long long clio_realtime_now_ns() {
+  return static_cast<long long>(
+      std::chrono::duration_cast<std::chrono::nanoseconds>(
+          std::chrono::system_clock::now().time_since_epoch())
+          .count());
+}
+
 /* Identity + state of the native file, as a string. dev/ino catch the file
    being replaced (a new inode at the same path -- h5repack, rsync, mv); size
    and mtime catch it being modified in place.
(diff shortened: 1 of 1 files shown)


hyoklee:
Verified the macOS branch on a real runner, since this repo's macOS VOL leg (`adapters (macos, hdf5_vol + macfuse)`) is gated `if: github.event_name != 'pull_request'` and skips here.

Built this branch at `e3f32dd7` on `macos-26` against HDF5 `develop` (2.3.0), via a build-only probe in hyoklee/hpf that publishes nothing ([run 31663063687](https://github.com/hyoklee/hpf/actions/runs/31663063687)):

```
[100%] Linking CXX shared library ../../../bin/libclio_hdf5_vol.dylib
[100%] Built target clio_hdf5_vol
libclio_hdf5_vol.dylib -> libhdf5: @rpath/libhdf5.1000.dylib
libclio_hdf5_vol.dylib resolves libhdf5.1000.dylib -> .../hdf5-install/lib/libhdf5.1000.dylib
```

No `st_mtim` diagnostics. On `dev` the same probe configuration fails with four `error: no member named 'st_mtim' in 'stat'`.

The Windows leg of that probe is re-running now (its first attempt died in netCDF-C's configure for want of `m4`, before it reached the connector — a gap in the probe, not in this branch). I'll post the MSVC result here when it lands.

hyoklee:
Windows verified too, same branch at `e3f32dd7`, MSVC 19.4x on `windows-2025` ([run 31663577228](https://github.com/hyoklee/hpf/actions/runs/31663577228)):

```
clio_hdf5_vol.vcxproj -> D:\a\hpf\hpf\nc4-clio-work\clio-build\bin\clio_hdf5_vol.dll
```

No `C2039: 'st_mtim'`, no `C2065: 'CLOCK_REALTIME'`, no `C3861: 'clock_gettime'` — and no other diagnostics took their place, so those really were the only non-portable uses in the file. That is the first time `clio_hdf5_vol.dll` has built here; the same probe on `dev` fails with those six errors.

So all three platforms are now covered for this change:

| platform | result |
| --- | --- |
| Linux (glibc 2.39, gcc 13) | builds; netCDF-4 workload through the connector produces the same 18 timings and the same stamp text as before |
| macOS 26 (clang) | `libclio_hdf5_vol.dylib` builds |
| Windows 2025 (MSVC) | `clio_hdf5_vol.dll` builds |

Worth considering separately: `ci-adapters.yml` gates both off-Linux VOL legs on `github.event_name != 'pull_request'` and `ci-windows.yml` configures `-DCLIO_CTE_ENABLE_HDF5_VOL=OFF`, so no pull request can catch a portability break in this file — which is how this one reached `dev`. A compile-only VOL job on macOS and Windows for PRs touching `adapter/hdf5_vol/` would have caught it in minutes, without the cost of the full adapter matrix. Happy to open that as a follow-up if you want it.

PR #975, opened 2026-08-13
Build pip and conda packages against glibc 2.28 so RHEL8 works (#973)

### LINKED ISSUE #973: iowarp-core wheels missing for manylinux_2_28 (glibc 2.28, e.g. Rocky Linux 8 HPC nodes)
iowarp-core (confirmed on 2.1.0 and 2.2.0, likely all releases ≥1.0.1) publishes wheels only for manylinux_2_34 (x86_64/aarch64), plus macOS/Windows — no manylinux_2_28 wheel and no sdist. This makes it impossible to install on glibc 2.28 hosts, e.g. Rocky Linux 8.10 (a common CHPC/HPC login-node OS), even though clio-agent's pyproject.toml only requires iowarp-core>=2.1.0 with no platform restriction.

Repro:
$ uv venv --python 3.13
$ uv pip install iowarp-core==2.2.0
× No solution found when resolving dependencies:
  ╰─▶ Because iowarp-core==2.2.0 has no wheels with a matching platform tag
      (e.g., `manylinux_2_28_x86_64`) and you require iowarp-core==2.2.0, we
      can conclude that your requirements are unsatisfiable.
  hint: Wheels are available for iowarp-core (v2.2.0) on the following platforms:
  manylinux_2_34_aarch64, manylinux_2_34_x86_64, macosx_14_0_arm64, win_amd64

Environment: Rocky Linux 8.10 (Green Obsidian), glibc 2.28 (ldd (GNU libc) 2.28), x86_64, uv 0.11.14.

Ask: publish a manylinux_2_28-compatible wheel (or an sdist that builds without requiring glibc 2.34+) so clio-agent can run on RHEL8-family HPC nodes.

lukemartinlogan:
Fixes #973 — `iowarp-core` is uninstallable on the RHEL8 family (Rocky/Alma/RHEL 8, glibc 2.28), the standard HPC login-node OS. Both distribution channels were broken, for two different reasons.

## pip (the reported bug)

Wheels were built in `manylinux_2_34` (AlmaLinux 9) and tagged to match, so pip on a glibc-2.28 host reports "no wheels with a matching platform tag" — and there is no sdist to fall back to.

Build in `manylinux_2_28` instead. This is a strict widening: a 2_28 wheel installs on glibc >= 2.28, covering RHEL8 *and* every distro the 2_34 wheels already covered. `installers/pip/README.md` already documented `manylinux_2_28`, so this realigns CI with the documented intent.

## conda (found while checking whether the same issue existed — it does, and it's worse)

I pulled the published 2.2.0 package off anaconda.org and inspected it: `libclio_run_cxx.so` imports `GLIBC_2.38` symbols (`__isoc23_strtol` and friends), and 157 ELF files in the package are affected.

The recipe declared `{{ compiler('cxx') }}` but not `{{ stdlib("c") }}`. Modern conda-forge compiler packages no longer carry a sysroot of their own, so the build compiled against the CI runner's `/usr` headers (ubuntu-latest, glibc 2.39). And because no `__glibc` constraint was recorded either, conda **installs it on Rocky 8 anyway** and the failure only surfaces at import — a silent failure rather than a clean solver error.

Fixed with `{{ stdlib("c") }}` + `c_stdlib_version: 2.28`. `conda-build render` confirms the build env now gets `sysroot_linux-64 2.28` and the package records `__glibc >=2.28,<3.0.a0`.

## Why nothing caught either one

Nothing in either pipeline connected the binaries actually produced to the glibc floor being advertised. `repair_wheel.sh` stamps the platform tag with `python -m wheel tags`, which rewrites metadata and never opens an ELF file (auditwheel, which *would* check, is deliberately not used). `conda build` doesn't check either. That gap is what let both defects ship.

This PR adds `installers/pip/check_wheel_glibc.py`: it unpacks a wheel or `.conda`, reads the versioned symbol requirements out of every ELF file, and fails if any needs a newer glibc than the package promises. It runs in both publish workflows before upload, and in the local `Dockerfile.build`.

It catches drift in both directions — a tag too new (this bug: users locked out) and a tag too old (wheel installs, then dies with `version 'GLIBC_2.34' not found`).

## Two incidental bugs found while verifying

- **`build_deps_manylinux.sh` used `curl --retry-all-errors`**, which requires curl >= 7.71. AlmaLinux 8 ships 7.61, where the unknown option makes curl exit immediately — so **every dependency download failed, on every retry**. Only caught by actually running the build in the new image. The wrapper's own retry loop already covers what the flag did.
- **`test_wheel.sh` reported `SKIP: No compatible wheel` for any install failure**, including real ones. It installed by name under `--no-index`, which fails both when no wheel matches the interpreter *and* when a runtime dep (pyyaml, flask, msgpack) can't be fetched — reporting both as a green skip. Now selects the wheel by path with the index available for runtime deps, as CI does.

## Other changes

- `repair_wheel.sh` takes the manylinux tag as an argument, passed from `matrix.cibw_image`, so the tag can't drift from the build image.
- `test-wheels` matrix gains `rockylinux:8` (x86_64 and aarch64) as the regression test, installing python3.12 from AppStream since the stock python3 is 3.6.
- README documents the actual floor: glibc 2.28+.

## Verification

Not just CI-config edits — this was built and run end to end locally:

- Built the full wheel in `manylinux_2_28`, FUSE adapter included (against el8's fuse3 3.3.0). Max glibc required: **2.27** across 29 ELF files.
- `iowarp_core-2.2.0-cp312-cp312-manylinux_2_28_x86_64.whl` installs on **rockylinux:8 (glibc 2.28)** and passes import / CLI / visualizer / symbol-resolution checks. Also passes on ubuntu:24.04 (no regression for existing users).
- The new checker was confirmed to **reject** the published 2.2.0 conda package, and to reject the same good wheel relabelled `2_17` (`memfd_create` pins the true floor at 2.27, so 2.28 is near-minimal).
- `conda-build render` verified for the recipe change.

## Not addressed here

The RPM/DEB/AppImage artifacts in `build-packages.yml` are built on Fedora 40 / Ubuntu 24.04 and also won't run on RHEL8. The distro packages are inherently distro-targeted; the AppImage claims portability and would need its own older build base. Worth a separate issue.

### DIFF
--- a/.github/workflows/build-pip.yml
+++ b/.github/workflows/build-pip.yml
@@ -57,6 +57,17 @@ jobs:
 
   #------------------------------------------------------------
   # Build manylinux wheels via cibuildwheel
+  #
+  # Image choice (#973): manylinux_2_28 (AlmaLinux 8, glibc 2.28), NOT
+  # manylinux_2_34 (AlmaLinux 9, glibc 2.34).  A manylinux_2_34 wheel is
+  # uninstallable on the RHEL8 family -- Rocky/Alma/RHEL 8 ship glibc 2.28
+  # and are still the standard HPC login-node OS, so `pip install
+  # iowarp-core` failed there with "no wheels with a matching platform tag"
+  # and there is no sdist fallback.  manylinux_2_28 wheels install on
+  # glibc >= 2.28, i.e. RHEL8 *and* every newer distro the 2_34 wheels
+  # already covered, so this is a strict widening.  Do not bump this image
+  # without a matching bump in installers/pip/repair_wheel.sh (which stamps
+  # the platform tag) and a reason to drop RHEL8.
   #------------------------------------------------------------
   build-wheels:
     runs-on: ${{ matrix.runner }}
@@ -68,7 +79,7 @@ jobs:
           - runner: ubuntu-latest
             arch: x86_64
             cibw_build: "cp310-manylinux_x86_64 cp311-manylinux_x86_64 cp312-manylinux_x86_64 cp313-manylinux_x86_64"
-            cibw_image: manylinux_2_34
+            cibw_image: manylinux_2_28
             cibw_environment: >-
               CMAKE_PREFIX_PATH=/usr/local
               PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/local/lib64/pkgconfig
@@ -77,7 +88,7 @@ jobs:
           - runner: ubuntu-24.04-arm
             arch: aarch64
             cibw_build: "cp310-manylinux_aarch64 cp311-manylinux_aarch64 cp312-manylinux_aarch64 cp313-manylinux_aarch64"
-            cibw_image: manylinux_2_34
+            cibw_image: manylinux_2_28
             cibw_environment: >-
               CMAKE_PREFIX_PATH=/usr/local
               PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/local/lib64/pkgconfig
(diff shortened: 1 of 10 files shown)


lukemartinlogan:
Superseded: opening a full dev -> main release PR instead, which carries this fix along with the rest of dev.

PR #981, opened 2026-08-14
vfd: coalesce vector I/O instead of one pread/pwrite per element

### LINKED ISSUE #980: HDF5 VFD: vector I/O issues one pwrite + one CFS populate per element (~200x vs sec2 on strided contiguous-dataset writes)
## Summary

`H5FD__clio_write_vector` / `H5FD__clio_read_vector` issue **one `pwrite`/`pread` plus one CFS cache-populate round-trip per vector element**. For a strided selection on a *contiguous* (unchunked) dataset, HDF5 hands the driver tens of thousands of 4-byte elements in a single call, so the VFD performs ~200x worse than `sec2` on the same operation.

The irony is that the VFD is penalised **because** it implements vector I/O: implementing the vector callbacks switches HDF5 to selection I/O, which bypasses the library's data-sieve buffer that `sec2` benefits from.

## Reproducer

netCDF-C `main` `tst_chunks3` (from `nc_perf`), single node, HDF5 `develop` `e4b6a96`, clio-core `dev` `edd8ab45`:

```
HDF5_DRIVER=clio_vfd HDF5_DRIVER_CONFIG="cache=1" \
  tst_chunks3 6 256 64 256 64 256 64
```

The timing of interest is `contiguous write 256 256 1` — an unchunked 256³ float variable written one xy-slab at a time.

| variant | `contiguous write 256x256x1` |
| --- | ---: |
| sec2 (baseline) | **6.1 s** |
| `clio_vfd` | **1200 s** (~197x) |

At 512³ the same operation does not finish in 90 minutes.

## What the driver receives

Breaking on `H5FD__clio_write_vector` during that operation (arguments read from the SysV registers; the Release build has no DWARF):

```
VEC count=65536
  addrs[0..2]: 0x1fb0  0x23b0  0x27b0      <- stride 0x400 = 1024 = N*sizeof(float)
  sizes[0..2]: 0x4     0x4     0x4
```

So HDF5 passes **the whole 256x256 slab as one vector of 65,536 four-byte elements**, and `H5FD__clio_write_vector` loops over them one at a time. `H5FD__clio_write` (the scalar entry) is never called.

`strace -c` over 5 s inside this operation, same operation and dimensions (512³) for both drivers:

| | calls | size of each |
| --- | ---: | --- |
| sec2 | 22,697 `pwrite64` + 22,698 `pread64` | **65,536 B** |
| `clio_vfd` | 19,535 `pwrite64` | **4 B** |

Per 15 s the VFD also makes 393,016 `getpid` and 361,229 `gettid` (~13 identity syscalls per element) and roughly one `tgkill` per element — the cost of the per-element `CLIO_CFS_CLIENT->PwriteFd()` round-trip in `H5FD__clio_do_write`.

## Root cause

HDF5 decides per I/O whether to use selection I/O (`H5Dio.c`, `H5D__ioinfo_adjust`):

```c
/* If selection I/O mode is default (auto), enable it here if the VFD supports it
   ... otherwise disable it */
if (io_info->use_select_io == H5D_SELECTION_IO_MODE_DEFAULT) {
    if (H5F_has_vector_select_io(..., op_type == H5D_IO_OP_WRITE))
        io_info->use_select_io = H5D_SELECTION_IO_MODE_ON;
    else
        io_info->use_select_io = H5D_SELECTION_IO_MODE_OFF;
}
```

* `H5FDsec2.c` registers `NULL /* read_vector */, NULL /* write_vector */`. Selection I/O is therefore **off**, the write goes through `H5D__contig_writevv`, and the **data sieve buffer** coalesces the strided elements into 64 KB read-modify-write blocks.
* `H5FDclio.cc` registers both vector callbacks. Selection I/O is therefore **on**, the sieve buffer is bypassed, and the raw element list reaches the driver.

`H5FD_FEAT_DATA_SIEVE` is advertised by the VFD but is inert once selection I/O is on: it governs a path the library is no longer taking.

The vector implementation's own comment anticipates the gap:

> Note the elements are still issued as individual pread/pwrites; coalescing file-contiguous elements into a single preadv/pwritev is a possible future optimization (worthwhile only for patterns with contiguous runs, which HDF5's scattered vector I/O rarely has).

For this pattern the assumption is inverted: the elements are *not* file-contiguous, but they are dense and regularly spaced — exactly what a sieve buffer exists to exploit. Coalescing only exactly-contiguous runs would not help here.

## Proposed fix

Coalesce inside `H5FD__clio_write_vector` / `H5FD__clio_read_vector`, which is possible because the whole selection arrives in one call:

1. **Contiguous runs** (`addrs[i] + sizes[i] == addrs[i+1]`) merge into a single `pwrite`/`pread`. Always a win, no read-modify-write needed.
2. **Near runs**: group consecutive elements while the spanning region stays under a cap, then service the group with one `pread` of the span, patch the elements into the buffer, and one `pwrite` of the span — i.e. do in the driver what HDF5's sieve buffer would have done. The span end is by construction the end of the group's last element, so nothing is written beyond what the unmerged path would have written.
3. **One CFS populate per merged span** instead of one per element, which is where most of the per-element cost is.

A 64 KB cap (matching HDF5's default `H5Pset_sieve_buf_size`) turns the 65,536 element writes above into ~1,024 read-modify-write pairs. The cap should be configurable through the existing driver config string (which already parses `key=value`), e.g. `sieve=65536`, with `sieve=0` restoring the current behaviour.

Alternative, if coalescing is not wanted: **do not register the vector callbacks**, which puts HDF5 back on the sieve path and makes this operation behave like `sec2` — at the cost of vectored efficiency elsewhere.

## Environment

* clio-core `dev` `edd8ab45`, HDF5 `develop` `e4b6a96`, netCDF-C `main` `db5a7f93`
* Linux 5.15, Intel Xeon Silver 4114, `CLIO_CTE_ENABLE_VFD=ON`, DRAM CTE tier
* Full measurements and profiling method: https://github.com/hyoklee/ares/blob/main/doc/20260814_claude_netcdf_test_results.md

hyoklee:
Fixes #980.

## The problem

HDF5 hands a driver that implements the vector callbacks **the whole selection in one call**. For a strided slab of a contiguous (unchunked) dataset that is tens of thousands of 4-byte elements — measured on netCDF-C `tst_chunks3` at 256³ by breaking on the entry point:

```
VEC count=65536
  addrs[0..2]: 0x1fb0  0x23b0  0x27b0      <- stride 0x400 = 1024 = N*sizeof(float)
  sizes[0..2]: 0x4     0x4     0x4
```

`H5FD__clio_write_vector` looped over them one at a time, so each element cost one `pwrite` **and** one CTE populate round-trip. `strace -c` over 5 s of that operation, same dimensions for both drivers:

| | calls | size of each |
| --- | ---: | --- |
| sec2 | 22,697 `pwrite64` + 22,698 `pread64` | 65,536 B |
| `clio_vfd` (before) | 19,535 `pwrite64` | **4 B** |

The irony is that the VFD was penalised **because** it implements vector I/O. HDF5 turns selection I/O on iff the driver supports it (`H5Dio.c`, `H5D__ioinfo_adjust`), and that takes away the library's data-sieve buffer — which is what `sec2`, with `NULL` vector callbacks, is getting those 64 KB blocks from. `H5FD_FEAT_DATA_SIEVE` is advertised by this driver but is inert once selection I/O is on: it governs a path the library is no longer taking.

## The fix

Do in the driver what the sieve buffer would have done. Consecutive elements are grouped while their spanning region stays within a window, and each group is serviced as one I/O:

* elements that exactly tile the span need no read — gather and write once;
* otherwise read the span, patch the elements in, write it back. The span ends at the last element's end, so nothing outside the range the unmerged path would have written is touched;
* **one CTE populate per group** instead of per element, which is where most of the per-element cost was.

The window is `sieve=<bytes>` in the driver config string, default 65536 to match HDF5's own `H5Pset_sieve_buf_size` default. `sieve=0` restores the previous element-at-a-time behaviour.

One implementation note for review: the `sizes[]`/`types[]` shorthand (a `0` / `H5FD_MEM_NOLIST` entry at `i > 0` meaning "reuse the last explicit value") is monotone, so the state is snapshotted at a group's first element and replayed over that group — the grouping loop walks a group twice, once to size it and once to copy it.

## Measurements

ares, Xeon Silver 4114, DRAM CTE tier, HDF5 `develop` `e4b6a96`, netCDF-C `main` `db5a7f93`. `tst_chunks3 6 256 64 256 64 256 64`, seconds of CPU:

| operation | sec2 | before | after | |
| --- | ---: | ---: | ---: | --- |
| `contiguous write 256x256x1` | 6.1 | **1200** | **24** | 50x faster |
| `contiguous read 256x256x1` | 3.5 | 39 | 4.3 | 9x faster |
| all 18 timings | 13.4 | 1245 | 34.5 | 36x faster |

The worst operation drops from ~197x the `sec2` baseline to 3.9x; the variant as a whole from 93x to 2.6x.

## Correctness

`tst_chunks3` reads its data back but never compares it, so it cannot catch a corrupted write. The gate used here is `h5diff` against a `sec2`-written file: **files written by sec2, by the original driver and by this branch are identical.**

Existing suites, built from this branch and run against it:

* `clio_cte_vfd_no_runtime_tests` — `PASS: the driver is complete without CLIO`
* `clio_cte_vfd_unit_tests` — `PASS: native write-through verified` (22 checks ok; check 23 skipped, `h5dump` not installed on the test node)

## What a reviewer should look at

* The read-modify-write of a group's spanning region assumes a single writer for that region. This is the same assumption HDF5's own sieve buffer makes, but it is worth a second opinion for SWMR.
* Coalescing is validated on one access pattern (dense, regularly strided, ascending) plus the equality check above. Groups whose elements are sparse within the window trade a larger tier write for fewer round-trips; the 64 KB cap bounds that, but the trade-off has not been swept.
* `H5FD__clio_read_vector` gets the same treatment; it is exercised by the read side of the numbers above but has had less scrutiny than the write path.

Full measurements, profiling method and the raw traces: https://github.com/hyoklee/ares/blob/main/doc/20260814_claude_netcdf_test_results.md


context-transfer-engine/adapter/vfd/H5FDclio.cc
```diff
@@ -185,11 +188,18 @@ extern "C" {
  * read tier makes them meaningful. */
 typedef struct H5FD_clio_fapl_t {
   hbool_t cache_enabled; /* populate the CTE cache tier (default on) */
+  size_t sieve_max;      /* vector-I/O coalescing window, bytes (0 = off) */
```
mattjala, line 201: `H5Pset_fapl_clio` should be updated to initialize this new field. Right now it starts off uninitialized

context-transfer-engine/adapter/vfd/H5FDclio.cc
```diff
@@ -977,6 +1000,116 @@ static herr_t H5FD__clio_write(H5FD_t *_file, H5FD_mem_t type, hid_t dxpl_id,
...
+  *end_out = end;
+  *payload_out = payload;
+  return j;
+}
+
+/* Scratch for one coalesced group. Bounded by the sieve window, so this is a
+ * 64 KiB allocation by default and it is reused for the whole vector call. */
+static herr_t H5FD__clio_vec_scratch(std::vector<char> *buf, size_t need) {
+  if (buf->size() < need) {
+    buf->resize(need);
```
mattjala, line 1164: This should be error-checked, and `H5FD__clio_vec_scratch` should propagate any errors instead of always succeeding.

context-transfer-engine/adapter/vfd/H5FDclio.cc
```diff
@@ -977,6 +1000,116 @@ static herr_t H5FD__clio_write(H5FD_t *_file, H5FD_mem_t type, hid_t dxpl_id,
...
+  size_t payload = st.size;
+  uint32_t j = i + 1;
+  if (cap > 0) {
+    while (j < count) {
+      H5FD_clio_vecst_t nxt = st;
+      H5FD__clio_vecst_step(&nxt, j, sizes, types);
+      if (addrs[j] < addrs[i]) break;                 /* not ascending */
+      const haddr_t e = addrs[j] + nxt.size;
+      if (e <= addrs[i]) break;                       /* degenerate */
+      if ((uint64_t)(e - addrs[i]) > (uint64_t)cap) break;
```
mattjala: It seems like this window cap is checking against the wrong value and failing to enforce the claimed 64 KiB bound

context-transfer-engine/adapter/vfd/H5FDclio.cc
```diff
@@ -234,9 +246,20 @@ static bool H5FD__clio_apply_config_str(hid_t fapl_id, H5FD_clio_fapl_t *fa) {
         return false;
       }
       fa->cache_enabled = on ? 1 : 0;
+    } else if (e.first == "sieve") {
+      errno = 0;
+      char *endp = nullptr;
+      const unsigned long long v = strtoull(e.second.c_str(), &endp, 0);
+      if (errno != 0 || endp == e.second.c_str() || (endp && *endp != '\0')) {
```
mattjala: It looks like a negative value (`-1`) would pass all of these checks and end up storing SIZE_MAX as the effective value.

context-transfer-engine/adapter/vfd/H5FDclio.cc
mattjala, line 521: Since the CLIO FAPL now has a field that can have a dangerously invalid value, we should also check the validity of the provided `size` value here.

mattjala (changes requested):

context-transfer-engine/adapter/vfd/H5FDclio.cc
hyoklee: Fixed in fc6bb43f. `H5Pset_fapl_clio` now starts from `H5FD_clio_fapl_default_g` and overrides only the field it exposes:

```c
H5FD_clio_fapl_t fa = H5FD_clio_fapl_default_g;
fa.cache_enabled = cache_enabled;
```

Assigning the default struct rather than naming fields keeps this correct for whatever field is added next — `H5Pset_driver` copies the struct verbatim onto the FAPL, so anything left unset here reaches `H5FD__clio_open` as stack garbage, and for `sieve_max` that garbage is the size of a scratch allocation.

context-transfer-engine/adapter/vfd/H5FDclio.cc
hyoklee: Fixed in fc6bb43f. `resize()` is now caught and converted into the driver's own failure path:

```c
try {
  buf->resize(need);
} catch (const std::exception &ex) {
  errno = ENOMEM;
  H5FD_CLIO_ERROR(("could not allocate the " + std::to_string(need) +
                   "-byte vector-I/O scratch buffer: " + ex.what()).c_str());
  return FAIL;
} catch (...) { ... }
```

Both call sites already checked the return (`if (H5FD__clio_vec_scratch(...) < 0) return FAIL;`), so the propagation is complete once the function stops lying about success. Worth being explicit about why it cannot just throw: these callbacks are reached from HDF5's C frames, and unwinding an exception through them is undefined.

context-transfer-engine/adapter/vfd/H5FDclio.cc
hyoklee: Correct, and it was reachable. The check tested the candidate element's own end instead of the span the group would have. Those differ whenever an element falls inside the span accumulated so far, which a *read* vector may legitimately contain: with a 4 KiB window, an 8 KiB element at 0 followed by a 16-byte element at offset 8 gave `e - addrs[i] == 24`, which passed, while the span stayed 8 KiB — so the scratch allocation was twice the window.

Fixed in fc6bb43f:

```c
const haddr_t nend = (e > end) ? e : end;
if ((uint64_t)(nend - addrs[i]) > (uint64_t)cap) break;
```

plus a guard on entry, since a first element that already exceeds the window cannot be coalesced around within it — it stays a group of one and is serviced directly, allocating nothing.

Pinned by new section 24 of `test_vfd_adapter.cc`, which drives `H5FDread_vector` directly (the library never emits these shapes) and asserts on the bound itself through a new exported `H5FDclio_vec_max_span_g`, after first asserting that coalescing happened at all so the check cannot pass vacuously. I confirmed it fails against the previous arithmetic (span 8192 under a 4096 window) and passes with the fix.

context-transfer-engine/adapter/vfd/H5FDclio.cc
hyoklee: Right — `strtoull` wraps a negative rather than rejecting it, and leaves `errno` untouched, so nothing downstream could tell `-1` from a deliberate `ULLONG_MAX`. Fixed in fc6bb43f by refusing the sign before parsing:

```c
const size_t first = sv.find_first_not_of(" \t");
if (first == std::string::npos || sv[first] == '-') { /* refuse */ }
```

and then bounding the value against a stated maximum before the narrowing cast:

```c
#define H5FD_CLIO_SIEVE_MAX_LIM ((unsigned long long)1 << 30)   /* 1 GiB */
if (v > H5FD_CLIO_SIEVE_MAX_LIM) { /* refuse */ }
fa->sieve_max = (size_t)v;
```

The bound matters for its own sake — the window sizes a per-call scratch buffer, so an absurd value is an OOM rather than a slow open — and it also makes the cast exact where `size_t` is 32-bit. Covered in section 22: `sieve=-1`, `sieve=1073741825`, `sieve=abc` and `sieve=` are all refused; `sieve=0`, `sieve=4096` and `cache=1;sieve=8192` accepted.

context-transfer-engine/adapter/vfd/H5FDclio.cc
hyoklee: Agreed — a driver-info block set with a raw `H5Pset_driver(fapl, id, &fa)` never goes through the config-string parser, so a struct an application filled in itself could still carry a nonsense window. Fixed in fc6bb43f: the check now lives just after `H5FD__clio_apply_config_str`, so it covers both sources at the one point every FAPL path converges on, and fails closed the way the config string does.

```c
if (!H5FD__clio_sieve_valid(fa.sieve_max)) {
  errno = EINVAL;
  H5FD_CLIO_ERROR("driver FAPL: the vector-I/O coalescing window "
                  "(sieve_max) exceeds the maximum of 1 GiB");
  return nullptr;
}
```

`H5FD__clio_sieve_valid` is the shared predicate (0 = coalescing off, through `H5FD_CLIO_SIEVE_MAX_LIM`), so the config string and the FAPL cannot drift apart on what counts as valid.

mattjala (approved):

Issue #710, opened 2026-07-08
Battletest clio_run stop


lukemartinlogan:
# Make clio_run stop reliable and robust
Upgrade the clio runtime stop function to be more reliable. This will be a multi-part change that will require additional testing and evaluation.

## Add a --force flag to clio_run stop and use that in jarvis.
Jarvis provides a PsshExec method to parallelize clio_run stop. 
We should add a new --force flag to clio_run stop to kill the runtime ungracefully.
--force a new flag we should use to indicate the runtime should die ungracefull using exit() immediately.  When issuing the StopRuntime task to terminate the runtime. The admin task it calls context-runtime\modules\admin\src\admin_runtime.cc StopRuntime function. This should only use pool query local to stop the runtime.

## Distributed Jarvis Unit Test
We should add a distributed docker CI to ensure that clio_run stop consistently kills the whole runtime. We should NEVER have a stale runtime artifact after this command.

PR #987, opened 2026-08-18
ci: make chimaera_stop/run_tests.sh executable

lukemartinlogan:
The **Docker Distributed Stop Test** has failed on every single run since the workflow was added (#710) — 7 failures, 1 cancelled, **zero successes**. It is red on `dev` right now.

## Cause

```
./context-runtime/test/integration/chimaera_stop/run_tests.sh: Permission denied
##[error]Process completed with exit code 126
```

`ci-docker-distributed.yml:79` invokes the script directly, but it was committed `100644`. Every other integration driver is `100755`:

```
100755  integration/alltoone/run_tests.sh
100644  integration/chimaera_stop/run_tests.sh   <-- this one
100755  integration/dashboard/run_tests.sh
100755  integration/distributed/run_tests.sh
100755  integration/expand/run_tests.sh
100755  integration/leader_elect/run_tests.sh
100755  integration/migrate/run_tests.sh
100755  integration/reconnect/run_tests.sh
100755  integration/recovery/run_tests.sh
```

It has the same `#!/bin/bash` shebang and documents `./run_tests.sh` as its usage, so the missing bit is an oversight rather than a convention.

Worth noting how the failure presented: exit 126 came *after* the ~70-second in-container build of `clio_run`, so the job burned its full build every run and then never executed a single assertion. Nothing about `clio_run stop` has ever actually been covered by this job — the workflow looked like it was doing work.

`entrypoint.sh` in the same directory stays `100644` deliberately: `docker-compose.yaml` invokes it as `/bin/bash <path>`, so its mode is irrelevant.

## Verification

This workflow has no `pull_request` trigger (`push: [main, dev]` + `workflow_dispatch`), so a PR run proves nothing. I dispatched it against this branch: https://github.com/iowarp/clio-core/actions/runs/32085162266

Since this job has never once passed, its assertions are entirely unproven — it is possible the test fails for a real reason once it can actually start. That result is the point of the dispatch, and it will be reported on this PR before merge.

## Related

`context-runtime/test/integration/collective_restart/run_tests.sh` is also `100644`, but no workflow references it, so it is left alone here.


lukemartinlogan:
## Dispatch result: the exec-bit fix works, and it uncovered the next bug

[Run 32085162266](https://github.com/iowarp/clio-core/actions/runs/32085162266) on this branch: **no `Permission denied`, no exit 126**. The script now executes, the 2-node cluster comes up, SSH is wired, and Jarvis initialises. So this PR does what it claims.

The job still fails, now for a real reason — the first assertion this test has ever reached:

```
Menu: pkg
Available commands:
  configure       Configure a package (aliases: pkg conf)
  readme          Show package README
  path            Show package directory paths
  help            Show package configuration help
Error: Package not found: runtime
```

`entrypoint.sh:74` calls:

```bash
jarvis pkg append jarvis_clio_core.clio_runtime runtime
```

but `jarvis pkg` has no `append` subcommand — it printed its menu and did nothing, so the following `jarvis pkg conf runtime` (line 75) failed with "Package not found: runtime". The repo's own docs use the pipeline form, `jarvis pipeline append <pkg> ...` (`jarvis_clio_core/*/USE.md`), which suggests `ppl append` is the intended spelling. `jarvis_clio_core/clio_adapters/README.md:165` also says `jarvis pkg append`, so that doc is stale in the same way.

This is exactly the class of bug a job that never starts will hide: #710 shipped an integration test whose Jarvis wiring has never once been executed.

**Scope note:** this PR deliberately stops at the exec bit. Fixing the Jarvis invocation is a separate change, and since the test has never run end to end there may be further breakage behind it — that work belongs with whoever owns #710 rather than being smuggled into a one-line mode change.

PR #998, opened 2026-08-18
ci(#710): make the distributed docker stop test actually run and pass

lukemartinlogan:
## Summary

The **Docker Distributed Stop Test** workflow (`ci-docker-distributed.yml`) has failed on every push to `dev` since it landed. It was failing at three successive layers, each masked by the one before it, plus the gate itself had a bug that made it non-distributed. All four are fixed here; the test now passes end-to-end locally with the workflow's exact build recipe.

### 1. jarvis-cd in the CI image is too old (the instant failure on dev)
`jarvis_clio_core.{clio_runtime,clio_cte,clio_cte_libfuse}` import `jarvis_cd.util.container_utils`, which only exists on jarvis-cd **`dev`**. `iowarp/deps-cpu:latest` ships a `main` checkout → `jarvis ppl append` dies with `No module named 'jarvis_cd.util.container_utils'`.

da63e60f (#989) already fixes the Dockerfile, but `build-dep-containers.yaml` only rebuilds/publishes the image on pushes to **`main`** (path-filtered) — last publish was 2026-05-21 — so `dev` stays red until then. `run_tests.sh` now clones jarvis-cd `dev` into `external/jarvis-cd` when absent (`JARVIS_CD_REPO` / `JARVIS_CD_REF` overrides); the compose file already puts that path first on `PYTHONPATH`. The test no longer depends on which jarvis-cd the deps image happens to carry. `external/jarvis-cd/` is gitignored.

### 2. Empty pipeline env → `clio_run: command not found` on both nodes
The imperative `jarvis ppl create` starts with an **empty** env, and `PsshExec` only forwards the pipeline env to the remote shells, so every pssh'd `clio_run` failed and the runtime never started. Added `jarvis ppl env build` (captures `PATH`/`LD_LIBRARY_PATH`). `jarvis ppl run yaml` does this implicitly, which is why the sibling `jarvis_deploy` test never hit it.

### 3. Admin ChiMod not built → `Admin ChiMod not found! This is a required system component.`
The workflow compiled only `clio_run`, but `clio_run runtime start` needs `libclio_admin_runtime.so` (discovered by scanning `LD_LIBRARY_PATH` = `build/bin`). Now builds `clio_run clio_admin_runtime`.

### 4. Gate only ever probed node1
`ssh` inside `while read host ... < "$HOSTFILE"` swallowed the rest of the hostfile on stdin, so `assert_all` silently checked node1 only — node2 was never asserted. `ssh -n` fixes it.

## Verification
Ran `run_tests.sh` locally against a build produced with the workflow's exact recipe (deps-cpu image, `cmake --preset release -DCLIO_CORE_ENABLE_PYTHON=OFF`, targets `clio_run clio_admin_runtime`):

```
[status:running] iowarp-jarvis-stop-node1 -> rc=0   clio runtime (port 9413): RUNNING (pid 81)
[status:running] iowarp-jarvis-stop-node2 -> rc=0   clio runtime (port 9413): RUNNING (pid 53)
[status:stopped] iowarp-jarvis-stop-node1 -> rc=1   clio runtime (port 9413): STOPPED (clean)
[status:stopped] iowarp-jarvis-stop-node2 -> rc=1   clio runtime (port 9413): STOPPED (clean)
=== Phase 2: force stop ... ===   (same, both nodes RUNNING → STOPPED (clean))
RESULT: PASS -- clio_run stop killed the whole runtime on every node, no stale artifacts
```

Note: the workflow only triggers on `push` to `main`/`dev` and `workflow_dispatch`, so it won't run on this PR automatically — dispatch it on this branch to see it green before merge, or it'll run on the merge to `dev`.

Refs #710. Follows #985/#987/#989.

### DIFF
--- a/.github/workflows/ci-docker-distributed.yml
+++ b/.github/workflows/ci-docker-distributed.yml
@@ -55,10 +55,13 @@ jobs:
       - name: Pull dependency image
         run: docker pull "$DEPS_IMAGE"
 
-      # Compile ONLY clio_run (its runtime start/stop/status subcommands are all
-      # this test needs) inside deps-cpu, so it links the same libraries the node
-      # containers provide at runtime.
-      - name: Build clio_run (inside deps-cpu)
+      # Compile ONLY clio_run plus the admin ChiMod inside deps-cpu, so they link
+      # the same libraries the node containers provide at runtime. clio_run's
+      # start/stop/status subcommands are all this test drives, but `runtime
+      # start` refuses to come up without libclio_admin_runtime.so ("Admin
+      # ChiMod not found! This is a required system component."), and the
+      # runtime discovers ChiMods by scanning LD_LIBRARY_PATH (build/bin here).
+      - name: Build clio_run + admin ChiMod (inside deps-cpu)
         run: |
           docker run --rm \
             -v "${{ github.workspace }}":/workspace -w /workspace -u 0 \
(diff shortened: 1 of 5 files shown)


lukemartinlogan:
Dispatched `ci-docker-distributed.yml` on this branch: **green** — https://github.com/iowarp/clio-core/actions/runs/32170905524 (jarvis-cd dev cloned, admin ChiMod built, both nodes RUNNING → STOPPED (clean) in graceful and --force phases, `RESULT: PASS`).

PR #1033, opened 2026-08-26
Give the Windows wheel/package test jobs the retry helper they call

lukemartinlogan:
Follow-up to #1031, which fixed the `--` binding bug in `CI/Retry.ps1`. That fix let `Test Windows wheels` actually run for the first time, and it failed in 11 seconds:

```
The argument 'CI/Retry.ps1' is not recognized as the name of a script file.
```

`test-wheels-windows` (build-pip.yml) and `test-windows` (build-packages.yml) both call `CI/Retry.ps1`, but neither job has a checkout step, so the helper is not on disk. This shipped with the helper itself and stayed invisible because the build job upstream failed first — neither test job ever reached its WinFsp step.

Neither job checks out the source **on purpose**: they install the built artifact and test it in isolation, and a repo tree in the working directory could shadow the very package under test. So this pulls in `CI/` and nothing else.

Two details that are easy to get wrong, both verified rather than assumed:

- **The step must run first, before the artifact download.** `actions/checkout` deletes the contents of a workspace that is not already a git repo — placing it after `Download Windows wheels` would delete the wheels it exists to help test. (My first attempt did exactly that.)
- **Non-cone mode is required.** Cone mode always adds the top-level files, `pyproject.toml` among them, which is precisely the shadowing these jobs avoid. Checked against a scratch repo: `git sparse-checkout set --no-cone CI` yields `CI/Retry.ps1` and `CI/retry.sh` alone — no root files, no source tree.

https://claude.ai/code/session_01U95XdU4TKKpSsDN9JJH4Nu

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -519,6 +519,20 @@ jobs:
     timeout-minutes: 45
     if: always() && needs.build-windows.result == 'success'
     steps:
+    # Runs FIRST, before any artifact download: this job has no source
+    # checkout by design (a repo working directory could shadow the extracted
+    # package), but it does need CI/Retry.ps1 -- and actions/checkout DELETES
+    # the contents of a workspace that is not already a git repo, so running it
+    # after "Download Windows package" would take the package with it.
+    #
+    # Non-cone mode is required: cone mode always adds the top-level files,
+    # pyproject.toml among them, which is exactly the shadowing this job avoids.
+    - name: Fetch the CI retry helper
+      uses: actions/checkout@v4
+      with:
+        sparse-checkout: CI
+        sparse-checkout-cone-mode: false
+
     - name: Download Windows package
       uses: actions/download-artifact@v4
       with:
--- a/.github/workflows/build-pip.yml
+++ b/.github/workflows/build-pip.yml
@@ -654,6 +654,20 @@ jobs:
     timeout-minutes: 30
 
     steps:
+    # Runs FIRST, before any artifact download: this job has no source
+    # checkout by design (a repo working directory could shadow the installed
+    # package), but it does need CI/Retry.ps1 -- and actions/checkout DELETES
+    # the contents of a workspace that is not already a git repo, so running it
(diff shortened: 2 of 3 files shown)


lukemartinlogan:
Closing as superseded — `dev` landed an equivalent fix for the same bug while this PR was open.

**What this PR fixed.** The Windows wheel/package test jobs called `pwsh CI/Retry.ps1 choco install winfsp -y --no-progress`, and `Retry.ps1` is an advanced function (`[CmdletBinding()]`), which makes every parameter positional in declaration order. `-Attempts` was declared first, so the bare `choco` bound to it:

```
Retry.ps1: Cannot process argument transformation on parameter 'Attempts'.
Cannot convert value "choco" to type "System.Int32".
```

**Why it's redundant.** `dev` fixed the same two problems a different way:

| | this PR (`d4df0df`) | `dev` |
|---|---|---|
| `Retry.ps1` binding | `[CmdletBinding()]` + `Position = 0` on `-Command` | `[CmdletBinding(PositionalBinding = $false)]` |
| helper availability | added a "Fetch the CI retry helper" step | `actions/checkout` with `sparse-checkout: CI` as the first step |

Superseded by a638f30a, 0d3f024a, and d35b140c.

Both variants are correct, and dev's is verified working — bare and named call shapes both bind, and the retry/backoff/exit-code path propagates the real code:

```
attempt 1/3 failed (rc=7); retrying in 1s
attempt 2/3 failed (rc=7); retrying in 2s
'cmd /c exit 7' failed 3/3 attempts (last rc=7)   -> exit 7
```

**Net contribution is zero.** Merging this branch with `dev` produces a tree byte-identical to `dev`:

```
merge tree:      4b8367fdbd0822b1a6d87f0901c3f58d86034237
origin/dev tree: 4b8367fdbd0822b1a6d87f0901c3f58d86034237
```

`git diff <merge> origin/dev` is empty across all files, so there is nothing left to merge. The two workflow-file conflicts that made this PR `CONFLICTING` are just the two branches having rewritten the same call sites.

Nothing is lost by closing — the fix is already on `dev`.

PR #1053, opened 2026-08-28
Give snapshot replica records their own entry tag

hyoklee:
## Summary
- `FlushMetadata` wrote **blob** records and **replica** records with the same `entry_type = 3`, so every replica record was parsed as a blob record.
- The reader's own replica branch (`else if (entry_type == 3)`) was therefore **unreachable dead code** — the blob branch above accepts `1|2|3`.
- Move replica records to `entry_type = 4` and point the reader's replica branch at 4.

## The bug

Writer, in `Runtime::FlushMetadata`:

| Location | Tag |
|---|---|
| `core_runtime.cc:5204` blob record | `uint8_t entry_type = 3;` |
| `core_runtime.cc:5269` replica record | `uint8_t rep_entry_type = 3;` ← collision |

Reader, in `Runtime::RestoreMetadataFromLog`:

| Location | Branch |
|---|---|
| `core_runtime.cc:5634` | `else if (entry_type == 1 \|\| entry_type == 2 \|\| entry_type == 3)` — claims 3 |
| `core_runtime.cc:5739` | `else if (entry_type == 3)` — **never reached** |

The writer's own comment says the replica record uses "a NEW type … so an old reader stops loudly instead of parsing replica blocks as the next entry." The intent was right; the value chosen was already taken by the blob record written immediately above it.

## Why it is not cosmetic

Parsing a replica record as a blob record desynchronises the stream — the reader continues, interpreting string bytes as binary fields. The wreckage identifies itself, because the resulting "bdev pool ids" decode to ASCII:

| value read as a pool id | bytes |
|---|---|
| 1936876912 | `"pers"` |
| 1650422899 | `"st_b"` |
| 774975024 | `"0.1."` |
| 1769173605 | `"ersi"` |

— fragments of `persist_blob_*` names and `<major>.<minor>.<name>` composite keys.

Those ids match no registered target, and the volatile-block filter **fails open**: it tests `if (tinfo && tinfo->persistence_level_ == kVolatile)`, so a block whose target cannot be resolved is *kept*. Blocks that a restore is supposed to drop therefore survive it.

Observed alongside this on a 50-blob snapshot: `RestoreMetadataFromLog: Restored 1 tags and 2 blobs` — 48 blobs lost to the desync.

## Compatibility

Types `1|2|3` still parse as blob records, so existing snapshots keep restoring their blobs.

Snapshots **already written** cannot be fully recovered: a type-3 record in them is genuinely ambiguous between a blob and a replica, and the format has no version header to disambiguate. This change stops *new* snapshots being written into that ambiguity, and makes replica layouts restore at all. An older reader meeting a type-4 record hits the existing `Unknown entry type` warning — the "stop loudly" behaviour the writer's comment intended.

## Test plan
- [x] Compiles clean (0 errors) on TACC Vista, GH200, Release + CUDA
- [x] Reader's replica branch is reachable for the first time (was dead code)
- [x] Applies cleanly to current `dev` (`d68f4f2b`); bug confirmed still present there
- [x] Full-suite run at the PR head (`b227b202`): **390/391**, the single failure being the
      pre-existing `cte_replication_persist_integration` (the placement bug fixed in #1052,
      which this PR does not claim to fix). 0 build errors. `cte_blob_replicas` and
      `cte_blob_replicas_force_net` both pass in-suite, with zero runtime-heap leak reports.

## What this does NOT fix

Found while investigating `cte_replication_persist_integration`. It is **not** that test's failure, and the test still fails identically with this applied — that one is a data-placement bug, fixed separately in #1052. This is landed on its own merits.

## Verification status

All items above are now done. Summary of what was measured, on TACC Vista
(GH200, CUDA 12.5, Release + CUDA) unless noted:

* **Full suite at the PR head:** 390/391; the one failure is pre-existing (#1052).
* **The fix discriminates:** the added restore test fails without the entry-tag
  change and passes with it, in both the plain and `force_net` variants.
* **Leak check:** verified on a `-DCLIO_CORE_ENABLE_LEAK_CHECK=ON` build — the
  config CI uses, and the only one where the assertion is compiled in at all.

### DIFF
--- a/context-transfer-engine/core/src/core_runtime.cc
+++ b/context-transfer-engine/core/src/core_runtime.cc
@@ -5252,11 +5252,24 @@ clio::run::TaskResume Runtime::FlushMetadata(clio::run::shared_ptr<FlushMetadata
       }
       task->entries_flushed_++;
 
-      // Entry type 3 == one replica's layout (issue #886), written right
-      // after its blob's type-2 record so restore can attach it to the
+      // Entry type 4 == one replica's layout (issue #886), written right
+      // after its blob's record so restore can attach it to the
       // just-inserted BlobInfo. A NEW type for the same no-version-header
       // reason as type 2: an old reader stops loudly instead of parsing
-      // replica blocks as the next entry. The snapshot MUST carry replicas —
+      // replica blocks as the next entry.
+      //
+      // It MUST NOT be 3, which is what it was until this fix: 3 is already
+      // the tag of the BLOB record written just above (droppable_ layout), and
+      // the reader's blob branch accepts 1|2|3. Every replica record was
+      // therefore parsed as a blob record, which desynchronised the whole
+      // stream: the reader went on to interpret string bytes as binary fields
+      // and produced garbage bdev pool ids (observed: 1936876912 == "pers",
+      // 1650422899 == "st_b"). Those ids matched no registered target, so the
+      // volatile-block filter -- which keeps a block when the target is
+      // unknown -- failed OPEN and the DRAM primary survived a reboot, which
+      // is exactly what cte_replication_persist_integration asserts against.
+      // The replica branch in the reader was dead code for the same reason.
+      // The snapshot MUST carry replicas —
       // FlushMetadata may truncate the WAL below, and the kExtendReplica
       // records being truncated are the only other place these layouts live.
       // Empty replicas are skipped; they hold nothing to restore and are
@@ -5266,7 +5279,7 @@ clio::run::TaskResume Runtime::FlushMetadata(clio::run::shared_ptr<FlushMetadata
         if (rep.blocks_.empty()) {
           continue;
         }
-        uint8_t rep_entry_type = 3;
+        uint8_t rep_entry_type = 4;
         uint32_t rep_idx = static_cast<uint32_t>(rep_i + 1);
         uint32_t rep_name_len = static_cast<uint32_t>(rep.name_.size());
         float rep_score = rep.score_;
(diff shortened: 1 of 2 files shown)


hyoklee:
## Verification + a note on the Codecov report

**Full suite (the check the description said was pending): 391/392.**

The single failure is `cte_replication_persist_integration`, which is the
pre-existing data-placement bug fixed separately in #1052 — it fails identically
on `dev` without this patch. So **this change introduces no regressions**.
Build clean, 0 errors, on TACC Vista (GH200, CUDA 12.5, Release + CUDA).

### The uncovered line

Codecov flags 1 of the 2 code lines as uncovered. Splitting them:

* the **writer** line (`rep_entry_type = 4`) is covered — `test_blob_replicas.cc`
  already snapshots a blob that has replicas;
* the **reader** line (`} else if (entry_type == 4) {`) is not.

That gap is the bug's own cause, not an oversight in this PR: the branch was
unreachable dead code precisely because **nothing restores a snapshot containing
replicas**. The only test that does is
`cte_replication_persist_integration`, which restarts a real runtime *process*
across two phases — and it is currently red for the unrelated placement bug in
#1052. Once that lands, this branch is exercised there.

I tried to close the gap with a unit test (put primary + replica →
`FlushMetadata` → `RestartContainers` → read the replica back). **It does not
work, and I am not proposing it.** `RestartContainers` inside one process does
not discard the in-memory metadata, so the post-restart read is served from live
state and never reaches `RestoreMetadataFromLog`. I confirmed this by A/B: the
test passes *unchanged with the fix reverted*, so it asserts nothing about the
restore path. Shipping it would have satisfied the line counter while proving
nothing.

Genuinely covering this line needs a process-level restart, i.e. an integration
test in the shape of `cte_replication_persist_integration`, not a unit test.
Happy to add one as a follow-up if reviewers want it gated on that rather than
inheriting coverage from #1052.

hyoklee:
## Coverage gap closed — correcting my earlier comment

My previous comment said covering the reader's replica branch needed a
process-level integration test. **That was wrong**, and this pushes the unit
test that does it.

The working approach is to compose a **second** CTE pool — new id, same
storage, same `metadata_log_path`, `restart_ = true` — rather than restarting
the existing one. A re-create of a live pool just returns it with its in-memory
metadata intact, so `RestoreMetadataFromLog` never runs; a cold second
container has to rebuild from the snapshot alone. That is the same path
`RestartContainers` drives after a reboot.

One detail that cost a round: the replica must be written `REPLICA_PERSISTENT`
so it lands on the file tier. A plain replica goes to the volatile RAM tier and
`RestoreMetadataFromLog` drops volatile blocks **by design** — that copy is
meant to die with the machine. Asserting a volatile replica survives a restore
is asserting the opposite of the intended behaviour.

### It discriminates

Same test, both ways, on TACC Vista (GH200, CUDA 12.5, Release + CUDA):

| | `cte_blob_replicas` | `cte_blob_replicas_force_net` |
|---|---|---|
| **without** the entry-tag fix | **FAILS** | **FAILS** |
| **with** it | Passes 4.06 s | Passes 3.49 s |

That pair is the point. My first attempt at this test passed with the fix
reverted — it would have turned Codecov green while exercising nothing — so a
green run on its own was not sufficient evidence to push it.

Standing results for the fix itself, unchanged: full suite **391/392**, the one
failure being the pre-existing placement bug fixed in #1052.

PR #1058, opened 2026-08-28
clio-vfd: do not attach a cache handle to a read-only open

### LINKED ISSUE #1056: clio-vfd: a read-only open attaches a CTE cache handle that nothing can ever use
## Summary

`H5FD__clio_open` attaches a CTE cache handle to files HDF5 opened **read-only**.
Nothing can ever use it: the tier is write-populated and not served on reads, and
the only writer cannot run on a read-only file. The handle is opened, never
touched, and closed — at the cost of a blocking `OpenFd` round trip on open and a
`Flush` plus `Close` on close.

## Where

`context-transfer-engine/adapter/vfd/H5FDclio.cc`

```c
  int fd = -1;
#if H5FD_CLIO_HAVE_CACHE_TIER
  if (fa.cache_enabled) {
    fd = CLIO_CFS_CLIENT->OpenFd(name, o_flags, H5FD_CLIO_POSIX_CREATE_MODE_RW);
  }
#endif
```

`fa.cache_enabled` is on by default and via `HDF5_DRIVER_CONFIG=cache=1`, and
`flags` is not consulted. But:

* `H5FD__clio_do_read` reads from `file->posix_fd` unconditionally — the driver's
  own comment says the tier is "populated on write, not yet served on reads".
* `H5FD__clio_do_write` is the only thing that writes to `file->fd`, and HDF5
  will not call it on a file opened without `H5F_ACC_RDWR`.
* `H5FD__clio_close` still pays `CloseFd` for it (`Client::CloseFd` in
  `filesystem/src/filesystem_client.cc`), which waits on an `AsyncClose` task.

So a read-only open/close cycle costs two blocking client round trips that
cannot change any result — and the `Open` one is serviced by
`Runtime::Open`, which for an open without `O_CREAT` awaits a `TagQuery` and
then a `GetTagSize`, two more CTE tasks inside the runtime.

## Impact

netCDF-C's `nc_test4/tst_files4` opens and closes the same file 32768 times:

```c
for (i = 0; i < 32768; i++) {
   if (nc_open(FILE_NAME, NC_NOWRITE, &ncid)) ERR_RET;
   if (nc_close(ncid)) ERR_RET;
}
```

That is ~66k client round trips of pure waste, each open one costing two further
CTE tasks in the runtime. The test finishes in 21.9 s on the stock HDF5 stack and
is killed at `ctest --timeout 1200` under `HDF5_DRIVER=clio_vfd`.

Published results: [NetCDF-C CLIO Tests — macOS](https://github.com/hyoklee/actions/wiki/NetCDF-C-CLIO-Tests-macOS).

## Suggested fix

Turn the cache off for a read-only open, decided before the
`H5FD__clio_cache_available()` probe so such an open also skips the runtime
attach and its retry timeout:

```c
  if (fa.cache_enabled && !(H5F_ACC_RDWR & flags)) {
    fa.cache_enabled = 0;
  }
```

Every tier operation in the driver is already guarded by
`H5FD__clio_cache_live(file->fd)`, which is false when no handle was opened, so
nothing else needs to change. Writable opens are unaffected.

When the tier does start serving reads, this becomes a decision to revisit — but
then it would be revisited with a reason, which is not the case today.

PR: TBD

hyoklee:
Fixes #1056.

`H5FD__clio_open` attaches a CTE cache handle whenever the FAPL asks for the
tier, without consulting `flags`. On a read-only file that handle can never be
used: `H5FD__clio_do_read` goes to the authoritative descriptor unconditionally
(the tier is "populated on write, not yet served on reads"), and
`H5FD__clio_do_write` — the only writer — cannot run without `H5F_ACC_RDWR`. The
handle is opened, never touched, and closed.

That is two blocking client round trips per open/close cycle, and the open one is
serviced by `Runtime::Open`, which for an open without `O_CREAT` awaits a
`TagQuery` and then a `GetTagSize` inside the runtime.

The fix turns the cache off for a read-only open, decided before the
`H5FD__clio_cache_available()` probe so such an open also skips the runtime
attach and its retry timeout. Every tier operation in the driver is already
guarded by `H5FD__clio_cache_live(file->fd)`, false when no handle was opened, so
nothing else changes. Writable opens are unaffected.

## Effect

netCDF-C's `nc_test4/tst_files4` does 32768 read-only `nc_open`/`nc_close` cycles
on one file — ~66k client round trips for a tier no byte is ever read from. On
macOS 12.7.6 Intel, 4 cores, `ctest --timeout 1200`:

| test | stock HDF5 | `dev` | this branch |
| --- | ---: | ---: | ---: |
| `nc_test4_tst_files4` | 21.9 s | timeout | **21 s** |

A loop of read-only `H5Fopen`/`H5Fclose` on a one-dataset file, which is that
test's inner loop isolated: **18.370 ms** per cycle on `dev`, **0.129 ms** here,
against 0.073 ms with no driver loaded.

## Full suite

The whole netCDF-C test suite, same build, one `clio_run` per variant,
`ctest -j4 --timeout 1200`:

| variant | result | wall clock |
| --- | --- | ---: |
| `baseline` (sec2 / native VOL) | 215/215 passed | 156 s |
| `clio_vfd` | **215/215 passed** | 618 s |
| `clio_vol` | **215/215 passed** | 1106 s |

No test regressed, and both CLIO variants now pass everything the stock stack
does. For comparison, the published macOS run has `clio_vfd` at 1 failure /
28m31s and `clio_vol` at 4 failures / 45m55s.

## Review note

When the tier does start serving reads this becomes a decision to revisit — but
then it would be revisited with a reason, which is not the case today.

### DIFF
--- a/context-transfer-engine/adapter/vfd/H5FDclio.cc
+++ b/context-transfer-engine/adapter/vfd/H5FDclio.cc
@@ -690,6 +690,25 @@ static H5FD_t *H5FD__clio_open(const char *name, unsigned flags,
     fa.fsync_on_flush = 1;
   }
 
+  /* A read-only open has nothing to gain from the cache tier, so do not attach
+     one. The tier is write-populated and not yet served on reads:
+     H5FD__clio_do_read goes to the authoritative descriptor unconditionally,
+     and H5FD__clio_do_write -- the only writer -- cannot run on a file HDF5
+     opened without H5F_ACC_RDWR. The handle would therefore be opened, never
+     used, and closed.
+     It is not free: the attach waits on a CFS Open task here and close() waits
+     on a Close, so a read-only open/close cycle pays two blocking round trips --
+     and the open one is serviced by Runtime::Open, which for an open without
+     O_CREAT awaits a TagQuery and then a GetTagSize inside the runtime. A
+     workload that opens and closes files in a loop spends its entire time there:
+     netCDF-C's nc_test4/tst_files4 does 32768 read-only open/close cycles, i.e.
+     ~66k client round trips for a tier no byte is ever read from.
+     Decided before the H5FD__clio_cache_available() probe below so a read-only
+     open also skips the runtime attach and its retry timeout. */
+  if (fa.cache_enabled && !(H5F_ACC_RDWR & flags)) {
+    fa.cache_enabled = 0;
+  }
+
   // Attach to the CLIO runtime ONLY when this file wants the cache tier: the
   // native file is authoritative, so the native-only configuration is a
   // complete driver on its own and must not require CLIO to be running.


hyoklee:
Re-verified against the exact commits that are failing in CI today, now that #1057 has merged.

Trees (all three are the SHAs the published macOS run #14 tested on 2026-08-31):

| source | commit |
| --- | --- |
| HDF5 `develop` | `904ef116924` |
| netCDF-C `main` | `61e3662f5` |
| clio-core `dev` | `523004c9` |

macOS 12.7.6, Intel, 4 cores, one `clio_run` per variant, `--timeout 1200`.

`nc_test4/tst_files4`, the only test still failing in that run:

| | result |
| --- | ---: |
| baseline (sec2 / native VOL) | 20 s, passes |
| `HDF5_DRIVER=clio_vfd`, `dev` as-is | **killed at 1200 s** (`rc=143`) |
| `HDF5_DRIVER=clio_vfd`, `dev` + this PR | **20 s, passes** |

The failure reproduces down to the shape of the captured log — output stops right
after `Admin client created`, because the test prints nothing during its 32768
read-only `nc_open`/`nc_close` iterations.

Whole netCDF-C suite on the same trees, `ctest -j4`:

| variant | result | wall clock |
| --- | --- | ---: |
| baseline | 215/215 passed | 72 s |
| `clio_vfd` + this PR | **215/215 passed** | 616 s |

So this is the last thing standing between the macOS matrix and green: #1057
took `clio_vol` from 4 failures to 0 (and 45m55s to 18m03s), and this PR takes
`clio_vfd` from 1 to 0.

The branch still applies to `dev` unchanged — I checked that the anchors around
the `fsync_on_flush` block and the `CLIO_CFS_CLIENT->OpenFd` attach site are
untouched by anything merged since, and GitHub reports the PR mergeable.

One note in case it is on anyone's mind: the HDF5 commit above includes
`Do not push an error when a driver has no info block`, which changes
`H5Pget_driver_info` — called directly in `H5FD__clio_open`. It makes no
difference to this failure in either direction; the timeout is the cache-handle
attach, not the FAPL lookup.

Issue #10, opened 2025-11-12
Remove elf dependency


hyoklee:
Macs will fail.
E.g., `this adds a FAKE package for osx.` 

Ref: https://anaconda.org/memfault/elfutils

PR #1063, opened 2026-08-29
ctp: stop charging the Windows commit limit for shared memory nobody touched

### LINKED ISSUE #1061: Windows: every shared-memory segment is fully committed against the paging file, so N clients cost N x 544 MB of commit before writing a byte
### Summary

On Windows, `SystemInfo::CreateNewSharedMemory()` backs every shared-memory segment with
the **paging file** and, because the section is `SEC_COMMIT` (the default), the whole size
is charged against the system commit limit the instant the section is created — before a
single byte is touched.

Neither POSIX branch of the same function behaves that way. Linux uses
`memfd_create()` + `ftruncate()`; macOS/BSD use a regular file plus `ftruncate()`. Both are
sparse and charge per page actually written.

`context-transport-primitives/src/system_info.cc`:

```cpp
#elif CTP_ENABLE_WINDOWS_SYSINFO
  std::string win_name = WinShmName(name);
  fd.windows_fd_ =
      CreateFileMappingA(INVALID_HANDLE_VALUE,   // use paging file
                         nullptr,
                         PAGE_READWRITE,
                         0,                         // high-order DWORD
                         static_cast<DWORD>(size),  // low-order DWORD
                         win_name.c_str());
  return fd.windows_fd_ != nullptr;
#endif
```

### Why it matters

`IpcManager::ClientInit()` creates a per-process data segment of
`memory.client_data_segment_size` plus `kShmMetadataOverhead` (32 MB). At the documented
default of 512 MB that is **544 MB of commit charge for every process that calls
`ClientInit`**, used or not — plus the runtime's own `main_segment_size` and
`runtime_data_segment_size`.

Measured on a 32 GB Windows 11 host (commit limit 37234 MB), creating eight 544 MB
segments each way and touching nothing:

| backing | segments created | commit charge | per segment |
| --- | ---: | ---: | ---: |
| `CreateFileMapping(INVALID_HANDLE_VALUE, …)` — today | 8 | **+4361 MB** | 545 MB |
| sparse file, `CreateFileMapping(hFile, …)` | 8 | **0 MB** | 0 MB |

The file-backed section still costs only what is written: after mapping a view, writing
the first bytes and the last page, commit rose by 3 MB.

### What it looks like when the limit is reached

Any workload that runs several CLIO clients at once exhausts the commit limit. Once it
does, the failures are spread across unrelated components and none of them names the
cause:

```
posix_shm_mmap.h:94 ERROR shm_init shm_open failed: Resource temporarily unavailable
ipc_manager.cc:2583 ERROR IncreaseClientShm: Failed to create shm for clio_980_0
ipc_manager.cc:429  ERROR ClientInit: Failed to create per-process shared memory
H5FDclio.cc:709   WARNING H5FD__clio_open CLIO runtime unavailable; opening x.nc native-only
```

```
Assertion failed: The paging file is too small for this operation to complete.
 (C:\vcpkg\buildtrees\zeromq\src\v4.3.5-0744e81c3a.clean\src\thread.cpp:57)
```

```
0 [main] sh 5049 sh.exe: *** fatal error - CreateFileMapping cygpid.5086, Win32 error 1455
0 [main] sh (2516) child_copy: cygheap read copy failed, ..., Win32 error 299
../test_common.sh: fork: retry: Resource temporarily unavailable
```

Win32 error 1455 is `ERROR_COMMITMENT_LIMIT`. Processes that never touch CLIO start
failing too, because the commit limit is a machine-wide resource.

Concretely, this is what the full netCDF-C test suite over the CLIO VFD and VOL on a
`windows-latest` runner looks like: 12 failures under `HDF5_DRIVER=clio_vfd` and 19 under
`HDF5_VOL_CONNECTOR=clio`, against 4 on the stock HDF5 stack — and every one of the 23
CLIO-only failures traces back to this. The proximate symptoms were an HDF5 chunk
allocation failing (`H5FL__malloc(): memory allocation failed for chunk`), a test's own
378 MB `malloc()` returning NULL, `STATUS_STACK_BUFFER_OVERRUN` aborts, and Git-Bash
being unable to fork.

### Two more defects in the same pair of functions

* **Sizes ≥ 4 GiB are silently truncated.** The high-order DWORD is hard-coded `0` and
  the low-order one is `static_cast<DWORD>(size)`, so a 4 GiB segment silently becomes a
  0-byte one and an 8 GiB one becomes 0 too. The call succeeds; the corruption shows up
  later.

* **`DestroySharedMemory()` is an empty branch on Windows.** Both POSIX branches
  `unlink()` the segment. `PosixShmMmap::shm_init()` calls it to clear a stale segment
  before creating a new one, so on Windows that call has never done anything.

### Proposed fix

Back the section with a sparse file in the existing per-user directory
(`GetMemfdDir()`, which already has a Windows branch), exactly as the macOS/BSD path
does, keeping the same `Local\<base>` section name so `OpenSharedMemory()` and
`SharedMemoryExists()` are untouched. `FILE_ATTRIBUTE_TEMPORARY` keeps the pages in
memory unless the system is genuinely short of it, so this is not a trip to the disk.
Pass the full 64-bit size, and make `DestroySharedMemory()` delete the file.

PR to follow.

### LINKED ISSUE #1062: shm creation failures are reported as strerror(errno), which is meaningless on Windows
### Summary

When a shared-memory segment cannot be created, `PosixShmMmap::shm_init()` reports the
reason as `strerror(errno)`:

`context-transport-primitives/include/clio_ctp/memory/backend/posix_shm_mmap.h`:

```cpp
    SystemInfo::DestroySharedMemory(url);
    if (!SystemInfo::CreateNewSharedMemory(fd_, url, backend_size)) {
      char *err_buf = strerror(errno);
      HLOG(kError, "shm_open failed: {}", err_buf);
      return false;
    }
```

Two things are wrong with that line, and both point a reader away from the cause.

**The function is not `shm_open()` on any platform.** It is `memfd_create()` on Linux,
`open()` on macOS/BSD and `CreateFileMappingA()` on Windows. Searching the codebase for
`shm_open` after seeing this message finds nothing.

**`errno` is meaningless on Windows.** The Win32 shared-memory calls report through
`GetLastError()` and do not touch `errno`, so what gets printed is whatever `errno`
happened to hold from some earlier, unrelated call. In practice that came out as:

```
posix_shm_mmap.h:94 ERROR shm_init shm_open failed: Resource temporarily unavailable
```

The real failure was `ERROR_COMMITMENT_LIMIT` (Win32 error 1455) — the machine was out of
commit charge. "Resource temporarily unavailable" is `EAGAIN`, which suggests a transient
condition worth retrying; the actual condition was neither transient nor retryable, and
was caused by this process asking for 544 MB it did not need (that is https://github.com/iowarp/clio-core/issues/1061, and
this message is why it took a while to find).

The message also does not say which segment or how big it was, which are the two facts
that identify the caller.

### Proposed fix

Add `SystemInfo::GetLastSharedMemoryError()`, returning `strerror(errno)` on POSIX and a
`FormatMessage`-rendered `GetLastError()` plus the numeric code on Windows, and report
the segment name and size alongside it:

```
could not create shared memory segment clio_980_0 of 570425344 bytes:
The paging file is too small for this operation to complete (Win32 error 1455)
```

PR to follow.

hyoklee:
Fixes #1061. Fixes #1062.

`SystemInfo::CreateNewSharedMemory()` backed every Windows segment with the paging file:

```cpp
CreateFileMappingA(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE,
                   0, static_cast<DWORD>(size), win_name.c_str())
```

A section created without `SEC_RESERVE` is `SEC_COMMIT`, so the whole size is charged against the system commit limit the moment the section exists, before a byte is written. Neither POSIX branch of the same function does that — Linux uses `memfd_create` + `ftruncate`, macOS/BSD a regular file, and both are sparse.

`IpcManager::ClientInit()` creates a per-process segment of `client_data_segment_size` + 32 MB, so at the default 512 MB that is **544 MB of commit charge for every process that calls `ClientInit`**, used or not.

### Measured

32 GB Windows 11 host, commit limit 37234 MB, eight 544 MB segments created each way, nothing written:

| backing | commit charge | per segment |
| --- | ---: | ---: |
| `CreateFileMapping(INVALID_HANDLE_VALUE, …)` — before | **+4361 MB** | 545 MB |
| sparse file, `CreateFileMapping(hFile, …)` — after | **0 MB** | 0 MB |

After mapping a view and writing the first bytes and the last page, the file-backed version had charged 3 MB. `OpenFileMappingA` by name still works and a second view reads back what the first wrote, last page included.

### Why it mattered

Once the commit limit is reached the failures land far from the cause: `CreateFileMapping` returns `ERROR_COMMITMENT_LIMIT`, ZeroMQ cannot create a thread ("The paging file is too small for this operation to complete"), HDF5 chunk allocations fail, and unrelated processes on the machine cannot fork. Running the full netCDF-C test suite over the CLIO VFD and VOL on a `windows-latest` runner, this accounted for **all 23 CLIO-only failures** — 12 under `HDF5_DRIVER=clio_vfd` and 19 under `HDF5_VOL_CONNECTOR=clio`, against 4 on the stock HDF5 stack.

### The change

Back the section with a sparse file in the per-user directory `GetMemfdDir()` already provides, as the macOS/BSD branch does. The section keeps its `Local\<base>` name, so `OpenSharedMemory()` and `SharedMemoryExists()` are untouched and cross-process attach by name is unchanged. `FILE_ATTRIBUTE_TEMPORARY` keeps the pages in memory unless the system is genuinely short of it, so this is not a trip to the disk. The section holds its own reference to the file, so the file handle is closed immediately and `File` stays a one-handle union.

Two more defects in the same pair of functions, folded in because they are the same three lines:

* **Sizes ≥ 4 GiB were silently truncated** — hard-coded `0` high-order DWORD plus `static_cast<DWORD>(size)`. Now passes the full 64-bit size.
* **`DestroySharedMemory()` was an empty branch on Windows**, left over from when there was no path to remove. Both POSIX branches unlink, and `PosixShmMmap::shm_init()` calls it to clear a stale segment before creating one, so on Windows that call had never done anything.

And #1062, because it is what made #1061 hard to find: `shm_init()` logged `"shm_open failed: {strerror(errno)}"`. The function is not `shm_open()` on any platform, and the Win32 calls do not set `errno`, so a commit-limit failure was reported as "Resource temporarily unavailable" — `EAGAIN`, which reads as transient and retryable, and was neither. New `SystemInfo::GetLastSharedMemoryError()` renders `GetLastError()` through `FormatMessage` on Windows and `strerror(errno)` on POSIX, and the message now names the segment and its size.

### Verification

Windows 11, MSVC 19.x (Visual Studio 18), vcpkg dependency set, `CLIO_CTE_ENABLE_VFD=ON CLIO_CTE_ENABLE_HDF5_VOL=ON` against HDF5 `develop`: builds clean, and the numbers above are from that build. The POSIX branches are untouched.

### DIFF
--- a/context-transport-primitives/include/clio_ctp/introspect/system_info.h
+++ b/context-transport-primitives/include/clio_ctp/introspect/system_info.h
@@ -232,6 +232,15 @@ class SystemInfo {
 
   CTP_DLL static void DestroySharedMemory(const std::string &name);
 
+  /** Why the last CreateNewSharedMemory/OpenSharedMemory call failed.
+   *
+   *  Callers used to report this as strerror(errno), which is right on POSIX
+   *  and meaningless on Windows: the Win32 shared-memory calls do not set
+   *  errno, so a commit-limit failure (ERROR_COMMITMENT_LIMIT) was reported as
+   *  whatever errno happened to hold -- in practice "Resource temporarily
+   *  unavailable", which points at the wrong problem entirely. */
+  CTP_DLL static std::string GetLastSharedMemoryError();
+
   CTP_DLL static void *MapPrivateMemory(size_t size);
 
   CTP_DLL static void *MapSharedMemory(const File &fd, size_t size, i64 off);
--- a/context-transport-primitives/include/clio_ctp/memory/backend/posix_shm_mmap.h
+++ b/context-transport-primitives/include/clio_ctp/memory/backend/posix_shm_mmap.h
@@ -90,8 +90,12 @@ class PosixShmMmap : public MemoryBackend, public UrlMemoryBackend {
 
     SystemInfo::DestroySharedMemory(url);
     if (!SystemInfo::CreateNewSharedMemory(fd_, url, backend_size)) {
-      char *err_buf = strerror(errno);
-      HLOG(kError, "shm_open failed: {}", err_buf);
+      // NOT strerror(errno): none of the three platform implementations is
+      // shm_open(), and the Win32 one does not set errno at all, so this used
+      // to report a commit-limit failure as "Resource temporarily
+      // unavailable". Ask the platform what actually went wrong.
+      HLOG(kError, "could not create shared memory segment {} of {} bytes: {}",
+           url, backend_size, SystemInfo::GetLastSharedMemoryError());
       return false;
     }
     url_ = url;
--- a/context-transport-primitives/src/system_info.cc
+++ b/context-transport-primitives/src/system_info.cc
@@ -603,20 +603,90 @@ bool SystemInfo::CreateNewSharedMemory(File &fd, const std::string &name,
   return true;
 #endif
 #elif CTP_ENABLE_WINDOWS_SYSINFO
+  // Back the section with a SPARSE FILE, not the paging file.
+  //
+  // The obvious spelling, CreateFileMapping(INVALID_HANDLE_VALUE, ...), asks
+  // for a pagefile-backed section -- and a section created without SEC_RESERVE
+  // is SEC_COMMIT, so the WHOLE size is charged against the system commit
+  // limit the moment it is created, before a single byte is touched. Neither
+  // POSIX branch above behaves that way: Linux's memfd_create + ftruncate and
+  // macOS's regular file are both sparse and charge per touched page.
+  //
+  // That asymmetry is not academic. A CLIO client's per-process data segment
+  // defaults to 512 MB (+32 MB metadata), so on Windows every process that
+  // called ClientInit charged 544 MB of commit whether it used any of it or
+  // not. Measured on a 32 GB host: eight such segments cost 4361 MB of commit
+  // with zero bytes written. Run a test suite that starts several client
+  // processes at once and the commit limit is reached, at which point
+  // CreateFileMapping fails with ERROR_COMMITMENT_LIMIT, ZeroMQ cannot create
+  // a thread, and unrelated processes on the machine start failing to fork.
+  //
+  // A file-backed section is charged to the file instead, and a sparse file
+  // allocates only the ranges written. The same eight segments then cost 0 MB
+  // of commit. FILE_ATTRIBUTE_TEMPORARY keeps the pages in memory unless the
+  // system is actually short of it, so this is not a trip to the disk.
+  //
+  // Everything else is unchanged: the section still carries the same
+  // "Local\<base>" name, so OpenSharedMemory/SharedMemoryExists are untouched,
+  // and a second process still attaches by name.
+  //
   // POSIX shared memory names start with `/` (e.g. "/ctp_shm_42"). Win32
   // kernel object names can't contain `/`, so map to "Local\<base>" — the
   // per-session namespace, which is right for single-host SHM. Without
   // this the mapping was created anonymously (nullptr name) and could
   // not be reopened by name, breaking every OpenSharedMemory.
   std::string win_name = WinShmName(name);
-  fd.windows_fd_ =
-      CreateFileMappingA(INVALID_HANDLE_VALUE,   // use paging file
-                         nullptr,                // default security
-                         PAGE_READWRITE,         // read/write access
-                         0,          // maximum object size (high-order DWORD)
-                         static_cast<DWORD>(size),  // low-order DWORD
-                         win_name.c_str());         // mapping object name
-  return fd.windows_fd_ != nullptr;
+
+  EnsureMemfdDir();
+  std::string shm_path = GetMemfdPath(name);
+
+  // FILE_SHARE_DELETE so DestroySharedMemory can unlink the path the way the
+  // POSIX branches do while a mapping is still live.
+  HANDLE hfile = CreateFileA(shm_path.c_str(), GENERIC_READ | GENERIC_WRITE,
+                             FILE_SHARE_READ | FILE_SHARE_WRITE |
+                                 FILE_SHARE_DELETE,
+                             nullptr, CREATE_ALWAYS,
+                             FILE_ATTRIBUTE_TEMPORARY, nullptr);
+  if (hfile == INVALID_HANDLE_VALUE) {
+    fd.windows_fd_ = nullptr;
+    return false;
+  }
+
+  // Sparse, so extending the file below reserves no disk either. Best effort:
+  // a filesystem without sparse support still gives a correct (if eagerly
+  // sized) file, and the commit-charge problem this function exists to avoid
+  // is solved by the file backing regardless.
+  DWORD unused = 0;
+  DeviceIoControl(hfile, FSCTL_SET_SPARSE, nullptr, 0, nullptr, 0, &unused,
+                  nullptr);
+
+  LARGE_INTEGER end;
+  end.QuadPart = static_cast<LONGLONG>(size);
+  if (!SetFilePointerEx(hfile, end, nullptr, FILE_BEGIN) ||
+      !SetEndOfFile(hfile)) {
+    CloseHandle(hfile);
+    fd.windows_fd_ = nullptr;
+    return false;
+  }
+
+  fd.windows_fd_ = CreateFileMappingA(
+      hfile,                                       // backing file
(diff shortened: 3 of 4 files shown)


hyoklee:
Verified against the workload that found this: the whole netCDF-C test suite (204 tests, `ctest -j4 --timeout 1200`) run three times over on Windows 11, one `clio_run` per variant, the same netCDF-C and HDF5 build for all three.

| variant | published CI run #10 (`windows-latest`) | with this PR |
| --- | --- | --- |
| `baseline` | 4 failed, 7 not run, of 204 | 204/204 passed |
| `clio_vfd` | 12 failed of 204 | **203/204 passed** |
| `clio_vol` | 19 failed of 204 | **203/204 passed** |

(The baseline row also needs [Unidata/netcdf-c#3446](https://github.com/Unidata/netcdf-c/pull/3446); `nc_perf` does not compile with MSVC, which cost 11 tests in every variant including the baseline. That is a separate problem from this one.)

The one remaining failure is the same test in both CLIO variants and is **not** this defect — see #1064. It reproduces identically on unpatched `dev`, which I rebuilt specifically to check that this PR was not the cause.

Two direct confirmations that the new path is doing what it says. With the patch, a running `clio_run` holds 3 675 MB of segments in `%TEMP%\clio_<user>\`:

```
chi_main_segment__9413        1024 MB   Valid Data Length is 0x0 (0)
chi_metadata_segment__9413    1024 MB   Valid Data Length is 0x0 (0)
chi_queue_segment__9413        472 MB   Valid Data Length is 0x0 (0)
clio_rambdev_..._301_0         513 MB   Valid Data Length is 0x0 (0)
clio_rambdev_..._512_1         513 MB   Valid Data Length is 0x0 (0)
clio-<pid>-shm-in-{0..3}     4x 32 MB   Valid Data Length is 0x0 (0)
```

Every one sparse, zero bytes of valid data, and zero commit charge. Before the patch that same startup charged all 3 675 MB against the commit limit — on a 16 GB `windows-latest` runner, roughly a quarter of the machine before a single test ran, with each of four concurrent test processes then asking for another 544 MB.

And per-process client segments (`clio_<pid>_<index>`) appear as files in that directory during a run, so the client path is exercised too; `OpenFileMappingA` by name still works and a second view reads back what the first wrote, last page included.

hyoklee:
Pushed `97db8412` for the `codecov/patch` failure (25%, target 50%).

The cause was straightforward: of the lines this PR adds, only two actually compile on Linux, and neither was reachable from any test.

* `SystemInfo::GetLastSharedMemoryError()` is new API and had no caller in the test suite at all.
* The rewritten failure report in `PosixShmMmap::shm_init()` only runs when a segment cannot be created, which nothing exercised.

Everything else in the diff is inside `#elif CTP_ENABLE_WINDOWS_SYSINFO`, so it is not compiled on the coverage runner and contributes no line records either way.

Two cases added to `test_system_info_exec`, which ctest already runs unconditionally as `ctp_system_info`:

| case | what it covers |
| --- | --- |
| `SystemInfoSharedMemoryError` | calls `GetLastSharedMemoryError()` cold and after a genuinely failed `OpenSharedMemory`, asserting it never falls through to the `"unknown error"` placeholder |
| `SystemInfoSharedMemoryCreateFailure` | drives the create failure with a 4096-character segment name, then checks `shm_init()` reports and returns false rather than mapping a backend it never created, and that `DestroySharedMemory()` on a segment that never existed stays a no-op now that the Windows branch actually deletes a file |

The long-name trigger is portable on purpose: `memfd_create(2)` caps the name at 249 bytes, and the macOS/BSD and Windows branches open a file whose path this makes far too long, so all three fail the create for their own reason.

Verified locally on Windows with MSVC 19.x — `All tests passed (40 assertions in 10 test cases)` — and the new message shows up in the run:

```
could not create shared memory segment xxx...xxx of 1048576 bytes:
The system could not find the environment option that was entered (Win32 error 203)
```

which is the change in one line: before this PR that read `shm_open failed: Resource temporarily unavailable`.

PR #1068, opened 2026-08-31
vcpkg: patch libzmq's Windows AF_UNIX signaler fallback (fixes #1067)

### LINKED ISSUE #1067: Windows: clients abort in libzmq's poller (epoll.cpp:73) during ClientInit
## Summary

On Windows, clio client processes abort inside libzmq during `IpcManager::ClientInit`, before the DEALER connects:

```
Bad file descriptor (C:\vcpkg\buildtrees\zeromq\src\v4.3.5-0744e81c3a.clean\src\epoll.cpp:73)
not a socket (C:\vcpkg\buildtrees\zeromq\src\v4.3.5-0744e81c3a.clean\src\epoll.cpp:73)
```

Both spellings are the same assert with a different `errno`.

## Diagnosis

`epoll.cpp:73` is the `errno_assert (rc != -1)` immediately after `epoll_ctl (EPOLL_CTL_ADD, ...)` in `zmq::epoll_t::add_fd`, so libzmq failed to register a descriptor with its poller. The logs place the abort precisely:

```
IpcManager::ClientInit: try_new_servers = 0
GetSharedContext [ZeroMqTransport] Created shared context with 8 I/O threads
not a socket (...epoll.cpp:73)                 <-- abort
IpcManager: DEALER transport connected...      <-- never reached
```

That window is I/O-thread startup, triggered by the first `zmq_socket()`, so the descriptor being registered is an I/O thread's mailbox signaler.

On Windows that signaler is an `AF_UNIX` socketpair created by `make_fdpair()` in libzmq's `src/ip.cpp`. `make_fdpair` keeps a tcp/ip fallback for systems where `AF_UNIX` does not work, but it clears that fallback the moment `bind()` succeeds, on the assumption that a successful bind means IPC works. A later `listen()` or `connect()` on the bound socket can still fail, and the function then returns -1 with no fallback left. `signaler_t`'s constructor does not check that return value, so `_r` stays `retired_fd`, `get_fd()` hands the invalid descriptor to the poller, and the poller aborts.

This is **zeromq/libzmq#4730**, fixed by **zeromq/libzmq#4734** (merged 2024-08-26). The fix is in no release: v4.3.5 (2023-10-09) is still libzmq's newest tag, so every vcpkg build of libzmq carries the bug. vcpkg's port already passes `-DZMQ_HAVE_IPC=0` for MinGW, which sidesteps it there; MSVC keeps IPC on.

Nothing is wrong in clio-core itself. clio-core exposes the bug by building a ZMQ context in every client process, so any Windows workload that spawns short-lived clients can hit it.

## Reproduction

Found through the netCDF-C test suite run against the CLIO VFD and VOL. `ncdump_tst_nccopy5` fails under both adapters and passes on the baseline:

| variant | result |
| --- | --- |
| `baseline` | 204/204 |
| `clio_vfd` | 203/204 — `ncdump_tst_nccopy5` |
| `clio_vol` | 203/204 — `ncdump_tst_nccopy5` |

It is deterministic rather than a load effect: run alone with `ctest -R ncdump_tst_nccopy5` on a runner that had executed nothing else, it still failed 1/1.

Two plausible causes were tested and **ruled out**:

* **I/O thread count.** `CLIO_ZMQ_IO_THREADS=1` reproduces the abort identically. Consistent with the diagnosis — the signaler pairs are `AF_UNIX` and consume no TCP ports at all.
* **Ephemeral port exhaustion.** The ports in flight at the failure were ~52240–52462 of the 49152–65535 range, and an exhausted range fails `bind()` with `WSAEADDRINUSE` rather than handing the poller a bad descriptor.

## Suggested fix

A vcpkg overlay port: `ports/zeromq` copied verbatim from the vcpkg registry plus the upstream patch, `port-version` bumped. `installers/vcpkg/overlay-ports` is already on `VCPKG_OVERLAY_PORTS` for every Windows build and already holds `libaec`, so no build-system change is needed. The upstream patch applies to v4.3.5 unmodified.

Verified: with the overlay port in place, the same isolated `ncdump_tst_nccopy5` run passes.

PR follows.

hyoklee:
Fixes #1067.

## The failure

On Windows, clio client processes abort inside libzmq during `IpcManager::ClientInit`, before the DEALER connects:

```
Bad file descriptor (C:\vcpkg\buildtrees\zeromq\src\v4.3.5-0744e81c3a.clean\src\epoll.cpp:73)
not a socket (C:\vcpkg\buildtrees\zeromq\src\v4.3.5-0744e81c3a.clean\src\epoll.cpp:73)
```

Both spellings are the same assert with a different `errno`.

## Why

`epoll.cpp:73` is the `errno_assert (rc != -1)` immediately after `epoll_ctl (EPOLL_CTL_ADD, ...)` in `zmq::epoll_t::add_fd`. The logs place the abort between the context being created and the DEALER connecting — I/O-thread startup, triggered by the first `zmq_socket()` — so the descriptor is an I/O thread's mailbox signaler.

On Windows that signaler is an `AF_UNIX` socketpair from `make_fdpair()` in libzmq's `src/ip.cpp`. `make_fdpair` keeps a tcp/ip fallback for systems where `AF_UNIX` does not work, but clears it as soon as `bind()` succeeds, assuming a good bind means IPC works. A later `listen()` or `connect()` can still fail, and the function then returns -1 with no fallback left. `signaler_t`'s constructor does not check that return value, so `_r` stays `retired_fd` and the poller aborts registering it.

That is **zeromq/libzmq#4730**, fixed by **zeromq/libzmq#4734** (merged 2024-08-26). The fix is in **no release** — v4.3.5 (2023-10-09) is still libzmq's newest tag — so every vcpkg build carries the bug. vcpkg's port already passes `-DZMQ_HAVE_IPC=0` for MinGW, sidestepping it there; MSVC keeps IPC on.

Nothing is wrong in clio-core. clio-core exposes the bug by building a ZMQ context in every client process, so any Windows workload spawning short-lived clients can hit it.

## The change

A vcpkg overlay port under `installers/vcpkg/overlay-ports/zeromq`:

* `ports/zeromq` copied verbatim from the vcpkg registry (4.3.5#3) — diff it against the registry port to confirm
* plus `windows-ipc-connect-fallback.patch`, the upstream diff, which applies to v4.3.5 unmodified
* `port-version` bumped 3 → 4 so the ABI hash changes and vcpkg rebuilds
* `README.md` recording why the port exists and the condition for deleting it (a libzmq release past v4.3.5)

`installers/vcpkg/overlay-ports` is already on `VCPKG_OVERLAY_PORTS` for every Windows build here and already holds `libaec`, so nothing in the build system changes.

## Verification

Found through the netCDF-C test suite run against the CLIO VFD and VOL, where `ncdump_tst_nccopy5` failed under both adapters and passed on the baseline:

| variant | before |
| --- | --- |
| `baseline` | 204/204 |
| `clio_vfd` | 203/204 — `ncdump_tst_nccopy5` |
| `clio_vol` | 203/204 — `ncdump_tst_nccopy5` |

The failure is deterministic, not a load effect: run alone via `ctest -R ncdump_tst_nccopy5` on a runner that had executed nothing else, it still failed 1/1. **With this overlay port, that same isolated run passes.**

Two candidate causes were tested and ruled out first:

* **I/O thread count.** `CLIO_ZMQ_IO_THREADS=1` reproduces the abort identically — consistent with the diagnosis, since the signaler pairs are `AF_UNIX` and consume no TCP ports.
* **Ephemeral port exhaustion.** Ports in flight at the failure were ~52240–52462 of the 49152–65535 range, and an exhausted range fails `bind()` with `WSAEADDRINUSE`, not a bad descriptor.

### Scope of what is verified

Directly verified on `clio_vfd`, via the isolated test. `clio_vol` failed the same test with the same assert, so the same fix should cover it, but I have not yet watched it pass — a full three-variant Windows run is in flight and I will post the numbers here when it lands.

### DIFF
--- a/context-transport-primitives/include/clio_ctp/introspect/system_info.h
+++ b/context-transport-primitives/include/clio_ctp/introspect/system_info.h
@@ -232,6 +232,15 @@ class SystemInfo {
 
   CTP_DLL static void DestroySharedMemory(const std::string &name);
 
+  /** Why the last CreateNewSharedMemory/OpenSharedMemory call failed.
+   *
+   *  Callers used to report this as strerror(errno), which is right on POSIX
+   *  and meaningless on Windows: the Win32 shared-memory calls do not set
+   *  errno, so a commit-limit failure (ERROR_COMMITMENT_LIMIT) was reported as
+   *  whatever errno happened to hold -- in practice "Resource temporarily
+   *  unavailable", which points at the wrong problem entirely. */
+  CTP_DLL static std::string GetLastSharedMemoryError();
+
   CTP_DLL static void *MapPrivateMemory(size_t size);
 
   CTP_DLL static void *MapSharedMemory(const File &fd, size_t size, i64 off);
--- a/context-transport-primitives/include/clio_ctp/memory/backend/posix_shm_mmap.h
+++ b/context-transport-primitives/include/clio_ctp/memory/backend/posix_shm_mmap.h
@@ -90,8 +90,12 @@ class PosixShmMmap : public MemoryBackend, public UrlMemoryBackend {
 
     SystemInfo::DestroySharedMemory(url);
     if (!SystemInfo::CreateNewSharedMemory(fd_, url, backend_size)) {
-      char *err_buf = strerror(errno);
-      HLOG(kError, "shm_open failed: {}", err_buf);
+      // NOT strerror(errno): none of the three platform implementations is
+      // shm_open(), and the Win32 one does not set errno at all, so this used
+      // to report a commit-limit failure as "Resource temporarily
+      // unavailable". Ask the platform what actually went wrong.
+      HLOG(kError, "could not create shared memory segment {} of {} bytes: {}",
+           url, backend_size, SystemInfo::GetLastSharedMemoryError());
       return false;
     }
     url_ = url;
--- a/context-transport-primitives/src/system_info.cc
+++ b/context-transport-primitives/src/system_info.cc
@@ -603,20 +603,90 @@ bool SystemInfo::CreateNewSharedMemory(File &fd, const std::string &name,
   return true;
 #endif
 #elif CTP_ENABLE_WINDOWS_SYSINFO
+  // Back the section with a SPARSE FILE, not the paging file.
+  //
+  // The obvious spelling, CreateFileMapping(INVALID_HANDLE_VALUE, ...), asks
+  // for a pagefile-backed section -- and a section created without SEC_RESERVE
+  // is SEC_COMMIT, so the WHOLE size is charged against the system commit
+  // limit the moment it is created, before a single byte is touched. Neither
+  // POSIX branch above behaves that way: Linux's memfd_create + ftruncate and
+  // macOS's regular file are both sparse and charge per touched page.
+  //
+  // That asymmetry is not academic. A CLIO client's per-process data segment
+  // defaults to 512 MB (+32 MB metadata), so on Windows every process that
+  // called ClientInit charged 544 MB of commit whether it used any of it or
+  // not. Measured on a 32 GB host: eight such segments cost 4361 MB of commit
+  // with zero bytes written. Run a test suite that starts several client
+  // processes at once and the commit limit is reached, at which point
+  // CreateFileMapping fails with ERROR_COMMITMENT_LIMIT, ZeroMQ cannot create
+  // a thread, and unrelated processes on the machine start failing to fork.
+  //
+  // A file-backed section is charged to the file instead, and a sparse file
+  // allocates only the ranges written. The same eight segments then cost 0 MB
+  // of commit. FILE_ATTRIBUTE_TEMPORARY keeps the pages in memory unless the
+  // system is actually short of it, so this is not a trip to the disk.
+  //
+  // Everything else is unchanged: the section still carries the same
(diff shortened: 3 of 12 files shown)


hyoklee:
Full three-variant Windows run against this branch: **204/204 in all three variants, zero failures, `ctest` exit 0 throughout.**

| variant | tests | passed | failed | wall clock | `ncdump_tst_nccopy5` |
| --- | ---: | ---: | ---: | ---: | --- |
| `baseline` | 204 | 204 | 0 | 150s | Passed (3.74s) |
| `clio_vfd` | 204 | 204 | 0 | 203s | **Passed (7.39s)** |
| `clio_vol` | 204 | 204 | 0 | 202s | **Passed (6.82s)** |

Built from:

| source | ref | commit |
| --- | --- | --- |
| HDF5 | `develop` | `904ef1169242` |
| netCDF-C | `main` | `61e3662f5cde` |
| clio-core | `fix/win-zmq-ipc-signaler` | `a97de3f07e5c` |

This closes the gap noted in the PR description. `clio_vol` was previously inference — it failed the same test with the same assert but I had not watched it pass. It now passes, so both adapters are directly verified, and no other test moved.

For comparison, the same suite immediately before this fix, on the same three sources apart from clio-core:

| variant | before | after |
| --- | --- | --- |
| `baseline` | 204/204 | 204/204 |
| `clio_vfd` | 203/204 | 204/204 |
| `clio_vol` | 203/204 | 204/204 |

Issue #1069, opened 2026-08-31
Windows: UnmapMemory never unmaps, leaking every view and breaking segment re-create


hyoklee:
## Description

`SystemInfo::UnmapMemory()` never frees anything on Windows. Both of its Win32 spellings are wrong:

```cpp
void SystemInfo::UnmapMemory(void *ptr, size_t size) {
#if CTP_ENABLE_PROCFS_SYSINFO
  munmap(ptr, size);
#elif CTP_ENABLE_WINDOWS_SYSINFO
  VirtualFree(ptr, size, MEM_RELEASE);
#endif
}
```

`UnmapMemory` is the counterpart of **two** different allocators, because on POSIX one `munmap` serves both:

* `MapSharedMemory()` → `MapViewOfFile` (`posix_shm_mmap.h` `_Detach`, `shm_deserialize`, `shm_init` error paths)
* `MapPrivateMemory()` → `VirtualAlloc` (`posix_mmap.h:127`)

On Win32 they need different deallocators, and `VirtualFree(ptr, size, MEM_RELEASE)` is wrong for each:

* A `MapViewOfFile` region must be released with `UnmapViewOfFile`. `VirtualFree` cannot free a `MEM_MAPPED` region and fails with `ERROR_INVALID_PARAMETER`.
* For a `VirtualAlloc` region, `MEM_RELEASE` **requires `dwSize` to be 0** ([VirtualFree docs](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree)). Passing the non-zero `size` fails with `ERROR_INVALID_PARAMETER`.

The return value is discarded, so both failures are silent. Every shared-memory view and every private mapping leaks for the lifetime of the process.

## Why it is visible now

It was harmless while Windows segments were pagefile-backed: there was no file to keep alive, and `CreateFileMappingA(INVALID_HANDLE_VALUE, ...)` on an existing name just opened the existing section.

Since #1063 backed segments with a sparse file, a leaked view keeps a live section **on that file**. `CreateNewSharedMemory` opens the backing file with `CREATE_ALWAYS`, which truncates — and truncating a file with a mapped section open is refused:

```
shm_init could not create shared memory segment cte_shm_cache_test_5432 of 67108864 bytes:
  The requested operation cannot be performed on a file with a user-mapped section open (Win32 error 1224)
```

## Steps to Reproduce

[...]

PR #1082, opened 2026-09-04
fix(ctp): actually unmap shared views on Windows

hyoklee:
## The failure

CDash [build 4029381](https://my.cdash.org/builds/4029381/tests) (site `win-10`, `dev/ninja/icx-2025.3.0`) had exactly one failing test: **`cte_shm_metadata_cache_all`**, 6 of 8 cases passing. The last two could not create their fixture:

```
shm_init could not create shared memory segment cte_shm_cache_test_7604 of 67108864 bytes:
  The system could not find the environment option that was entered (Win32 error 203)
  The requested operation cannot be performed on a file with a user-mapped section open (Win32 error 1224)
```

## Root cause

`SystemInfo::UnmapMemory`'s Windows branch was `VirtualFree(ptr, size, MEM_RELEASE)`. That is not the release call for either mapping the class hands out:

- `MEM_RELEASE` **requires** `dwSize == 0`, so a non-zero size fails with `ERROR_INVALID_PARAMETER`;
- `VirtualFree` cannot release a `MapViewOfFile` view at **any** size.

Nothing was ever unmapped on Windows — shared views or private `VirtualAlloc` memory. Confirmed with a standalone Win32 repro:

```
private VirtualFree(ptr,65536,MEM_RELEASE) -> 0 err=87     # ERROR_INVALID_PARAMETER
private VirtualFree(ptr,0,MEM_RELEASE)     -> 1 err=87     # succeeds
    VirtualFree(data,size,MEM_RELEASE) -> 0 err=87
    VirtualFree(hdr,size,MEM_RELEASE)  -> 0 err=87
    DeleteFileA -> 0 err=5
[2] CreateFileA -> FAIL err=1224                           # ERROR_USER_MAPPED_FILE
```

A leaked view only cost address space while the sections were pagefile-backed. Since fc7d738a ("stop charging the Windows commit limit for shared memory nobody touched") they are file-backed, so the chain bites: a live view pins the section, the section pins the backing file, `DestroySharedMemory`'s `DeleteFile` fails, and the next `CreateFile(CREATE_ALWAYS)` over that path fails with `ERROR_USER_MAPPED_FILE`.

`test_cte_shm_metadata_cache`'s fixtures all share one per-pid segment name, so only the first could ever create it. That commit landed 2026-08-28, between the last green Windows build and 4029381.

### The bogus "error 203"

A second, smaller bug made the first failure lie about its cause. `GetLastSharedMemoryError()` was passed as an argument to `HLOG`, and `HLOG`'s first use in a process constructs the `Logger` singleton, which looks up `CTP_LOG_LEVEL` and `CTP_LOG_OUT` and leaves `ERROR_ENVVAR_NOT_FOUND` in the thread's last-error. Evaluation order let that race the read, so the *first* shm failure of every process reported "could not find the environment option that was entered" while the second (logger already built) reported the true `ERROR_USER_MAPPED_FILE`.

## The fix

- `UnmapViewOfFile(ptr)` for views, falling back to `VirtualFree(ptr, 0, MEM_RELEASE)` for `MapPrivateMemory`'s `VirtualAlloc`. `UnmapViewOfFile` simply reports `ERROR_INVALID_ADDRESS` for anything that is not a view base, so the two cases separate cleanly.
- Read the error string into a local before logging it.

## Test plan

Two regression cases in `test_system_info.cc` build a segment of the same name up and down three times — one at the `SystemInfo` layer, one at the `PosixShmMmap` layer where the failure surfaced.

- Reproduced the exact CDash failure locally first (same site/toolchain).
- Both new cases **fail at round 1** with the fix reverted, with the same `ERROR_USER_MAPPED_FILE`; both pass with it.
- `cte_shm_metadata_cache_all` now passes 8/8.
- Full suite on win-10 / Ninja / icx-2025.3.0 Debug: **237/237 passing** (was 236/237).

Since the unmap now genuinely happens on Windows, the full-suite run also covers the risk that the old leak was masking a use-after-unmap somewhere; nothing regressed.

### DIFF
--- a/context-transport-primitives/include/clio_ctp/memory/backend/posix_shm_mmap.h
+++ b/context-transport-primitives/include/clio_ctp/memory/backend/posix_shm_mmap.h
@@ -94,8 +94,18 @@ class PosixShmMmap : public MemoryBackend, public UrlMemoryBackend {
       // shm_open(), and the Win32 one does not set errno at all, so this used
       // to report a commit-limit failure as "Resource temporarily
       // unavailable". Ask the platform what actually went wrong.
+      //
+      // Read it into a local BEFORE logging. GetLastError() is thread-global
+      // and clobbered by any intervening Win32 call, successful ones included
+      // -- and HLOG's first use in a process constructs the Logger singleton,
+      // which looks up CTP_LOG_LEVEL and CTP_LOG_OUT and so leaves
+      // ERROR_ENVVAR_NOT_FOUND behind. Passed as an argument to HLOG this
+      // raced the singleton's construction, and the first shm failure of every
+      // process was reported as "could not find the environment option that
+      // was entered" instead of its real cause.
+      std::string shm_err = SystemInfo::GetLastSharedMemoryError();
       HLOG(kError, "could not create shared memory segment {} of {} bytes: {}",
-           url, backend_size, SystemInfo::GetLastSharedMemoryError());
+           url, backend_size, shm_err);
       return false;
     }
     url_ = url;
--- a/context-transport-primitives/src/system_info.cc
+++ b/context-transport-primitives/src/system_info.cc
@@ -866,7 +866,33 @@ void SystemInfo::UnmapMemory(void *ptr, size_t size) {
 #if CTP_ENABLE_PROCFS_SYSINFO
   munmap(ptr, size);
 #elif CTP_ENABLE_WINDOWS_SYSINFO
-  VirtualFree(ptr, size, MEM_RELEASE);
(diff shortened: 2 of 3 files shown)


hyoklee:
Closing as a duplicate: #1069 already describes this bug and #1070 already fixes it, with a better discriminator (VirtualQuery + MEM_MAPPED) than the try-UnmapViewOfFile-and-fall-back used here. I missed the open issue/PR before writing this. Following up on #1070 with the one remaining piece from this branch — the `Win32 error 203` in the CDash log is a separate clobber that #1070 does not address.

